PR #4, opened 2015-12-08
Complete example code in Readme

Juice10:
I wasn't able to figure out what `link` and `hash` should be but I was able to finish a lot more of the code in the Usage part of the readme


mafintosh:
Thanks, it's much more clear now! 
I've added more info about the `link` variable in the example. The hash one was actually a typo.

PR #5, opened 2015-12-08
(WIP) complete 3rd usage script

Juice10:
Wasn't able to actually get this running correctly but was able to complete a bit of the script


mafintosh:
It is implied that this example expands on the one just before it. Maybe we should put a note with that in, instead of duplicating all the requires?

mafintosh:
we ended up changing the interface a bit

PR #7, opened 2015-12-10
add missing pump depends

ryanramage:
Just adding pump depends so `npm run bundle` works


mafintosh:
lol good catch thanks

PR #21, opened 2016-01-11
Allow "hypercore" as a valid protocol value

paulkernfeld:
This will make it easier to rename the protocol to hypercore later.

I'm basing this change on the assumption that eventually hyperdrive is supposed to depend on hypercore, which I'm not sure is true, so this pull request may be misguided.


mafintosh:
spot on :)

PR #25, opened 2016-01-30
prevents hanging when the download has already happened. fixes #24

### LINKED ISSUE #24: downloading twice never calls download callback
Create an archive and download the contents successfully, then exit the process. Then, run it again. You'll notice that the callback is never fired to finish the download. Example:

```
$ dat dat://.....
Download complete, etc.
^C
$ dat dat://
Connected to 1/1 peers
```

'download complete' never fires because the callback isn't called.

karissa:
this fixed the behavior but the tests aren't passing

### DIFF
--- a/index.js
+++ b/index.js
@@ -104,18 +104,16 @@ Archive.prototype.download = function (i, cb) {
     kick()
 
     function kick () {
-      if (!feed.blocks) return
       for (; ptr < feed.blocks; ptr++) {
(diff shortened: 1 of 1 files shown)


mafintosh:
fixed here, https://github.com/mafintosh/hyperdrive/commit/0c363e0f1845361c9d0ec118b99f06b30e06b132 :)

PR #26, opened 2016-02-11
Add stats object to Archive for progress reasons

joehand:
(wrote w/ Max)

### DIFF
--- a/README.md
+++ b/README.md
@@ -106,7 +106,7 @@ Emitted when a file is fully downloaded.
 
 #### `archive.append(entry, [opts], [callback])`
 
-Either returns a write stream if entry is a file or returns `false` if it is directory. Calls callback when the entry has finished writing. 
+Either returns a write stream if entry is a file or returns `null` if it is directory. Calls callback when the entry has finished writing. 
 
 #### `archive.appendFile(filename, [name], [callback])`
 
--- a/README.md
+++ b/README.md
@@ -104,6 +104,10 @@ Emitted when a data block if downloaded for a file.
 
 Emitted when a file is fully downloaded.
 
+#### `archive.append(entry, [opts], [callback])`
+
+Either returns a write stream if entry is a file or returns `false` if it is directory. Calls callback when the entry has finished writing. 
+
 #### `archive.appendFile(filename, [name], [callback])`
 
 Append a file to a non-finalized archive. If you don't specify `name` the entry will be called `filename`.
--- a/index.js
+++ b/index.js
@@ -369,6 +369,8 @@ Archive.prototype.appendFile = function (filename, name, cb) {
       size: 0,
       link: null
     }, {filename: filename}, function (err, entry) {
+      if (err) return cb(err)
+      if (!entry) return cb()
       if (entry.type === 'file') self.readStats.files++
       cb(err, entry)
     })
--- a/index.js
+++ b/index.js
@@ -7,6 +7,7 @@ var bulk = require('bulk-write-stream')
 var path = require('path')
 var deltas = require('delta-list')
 var from = require('from2')
+var through = require('through2')
 var pump = require('pump')
 var pumpify = require('pumpify')
 var octal = require('octal')
@@ -14,7 +15,7 @@ var util = require('util')
 var events = require('events')
 var storage
 var rabin
-if (process.browser) rabin = require('through2')
+if (process.browser) rabin = through
 else rabin = require('rabin')
 if (process.browser) storage = require('./lib/browser-storage')
 else storage = require('./lib/storage')
@@ -66,6 +67,7 @@ function Archive (drive, folder, id) {
   this.core = drive.core
   this.entries = 0
   this.metadata = {type: 'hyperdrive'}
+  this.readStats = {totalRead: 0, files: 0}
 
   this._first = true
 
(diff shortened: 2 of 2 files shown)


README.md
```diff
@@ -104,6 +104,10 @@ Emitted when a data block if downloaded for a file.
 
 Emitted when a file is fully downloaded.
 
+#### `archive.append(entry, [opts], [callback])`
+
+Either returns a write stream if entry is a file or returns `false` if it is directory. Calls callback when the entry has finished writing.
```
mafintosh: technically it returns `null` if it isn't a file

index.js
```diff
@@ -366,9 +368,23 @@ Archive.prototype.appendFile = function (filename, name, cb) {
...
+    var statsCounter = through(function (data, encoding, next) {
+      self.readStats.totalRead += data.length
+      this.push(data)
+      next()
+    })
+
+    if (ws) {
+      var readStream = fs.createReadStream(path.resolve(self.directory, filename))
+      pump(readStream, statsCounter, ws)
+    }
```
mafintosh: i'd rather return a stats object with `{read: increasingNumber, total: totalNumberOfBytes}` that only contains stats for this one append. the callback is already called with the entry so you can track the entry type elsewhere as well.

mafintosh:
@joehand cool stuff! i've added some comments :)

index.js
mafintosh: bytesRead, bytesTotal might be better names but i'll leave that up for you to decide.

joehand: @mafintosh that makes sense to return the stats for the single append rather than the whole archive. =)

I can get the appendFile to return the stats in the callback (the total append size is already there) but not while in progress on a single append (in the `through` stream). @maxogden, I may need your help here. The only way I could figure it out was to emit an event and use that in `Dat.addFiles`:

```javascript

Archive.appendFile = function(...) {
  ...
  var statsCounter = through(function (data, encoding, next) {
      this.push(data)
      self.emit('read-progress', data.length)
      next()
    })
  ...
}

Dat.addFiles = function(...) {
  var pack = this.drive.add('.')
  ...
  pack.on('read-progress', function(data) {
    stats.bytesRead += data
  })
  ...
}
```

Is there a better way monitor the read progress on a single file?

---

It seems like sometimes progress is tracked in both Hyperdrive and Dat, for different functions. A lot of the progress (read files, read bytes for finished files, etc.) can be either place, but some (e.g. reading progress on single file) can only be in hyperdrive. Should the goal be to keep as much progress stuff in dat except only hyperdrive can track things (e.g. `statsCounter`)? 

For instance, [this line](https://github.com/maxogden/dat/blob/master/index.js#L92) we moved from Dat to Hyperdrive but it is not quite clear to me which module it should be in.

maxogden: @joehand I think you'll need to return something immediately from `appendFile` that has the stats object attached to it. @mafintosh what do you think the API should be for that?

Also I think its OK hyperdrive has the stats tracking code in it, since other hyperdrive users beyond just dat would probably use it. I think dat should handle formatting the stats the printing them out, hyperdrive should just collect the stats as simply as possible and make them available. For things like averaging etc or doing the total sum of all files its prob easier to do it in dat right now.

maxogden: Oh yea was gonna mention we generally dont use events for stats cause events are relatively expensive, e.g. emitting 10000 events is a lot more expensive than updating an integer 10000 times

mafintosh: @joehand progress api:

``` js
var stats = drive.appendFile(..., function () {
  clearInterval(interval)
  console.log('File added')
})

var interval = setInterval(function () {
  console.log(stats.bytesRead + ' / ' + stats.bytesTotal)
}, 1000)
```

@maxogden yes i agree that a stats api is a good idea in general but if it tracks how many files were added it should track the other types as well, https://github.com/mafintosh/hyperdrive/blob/master/index.js#L377-L387.

mafintosh:
@joehand @maxogden merged the pr and tweaked the things i mentioned.

there is now a stats object on the feed containing

``` js
feed.stats = {
  bytesRead: bytesReadIntoTheFeed,
  bytesDownloaded: totalNumberOfBytesDownloaded
}
```

I wanted to add bytesUploaded as well but I don't expose all the information yet from hypercore. That'd be a cool addition though so I should fix that! Do you think this should expose readSpeed/downloadSpeed/uploadSpeed as well?

`.appendFile()` now returns a stat object

``` js
var stats = feed.appendFile(...)
...
console.log(stats) // {bytesRead, bytesTotal}
```
 
In regards to the file type tracking I wanted to expose that for archives that are being downloaded from remote sources as well but that turned out to be kinda tricky so I think we should just stick with tracking that in Dat for now using the type entry object returned in the callback.

mafintosh:
Released in 3.1.0 - let me know if you need more stats

index.js
mafintosh: @maxogden nvm my comment about tracking other types. i forgot we had the very file explicit api in there so that makes sense

mafintosh:
Actually just found a simple way to track files downloaded as well so thats in there now in addition to files read

``` js
feed.stats.filesRead // number of files that were read in appends
feed.stats.filesDownloaded // number of files that have been downloaded (in the current session)
```

Released in 3.2.0

joehand:
@mafintosh awesome, thanks! That should be good for what we were printing. 

@maxogden I'll update the PR on dat to use these stats.

PR #27, opened 2016-02-13
bytesRead progress for single append doesn't currently update

joehand:
Thanks for implementing the other stats progress. There was one bug, the bytesRead for a *single append* was stuck at 0, not updating. Currently only the total bytes read is updated (`Archive.stats.bytesRead`). 

This is one way to fix it but there may be a better way to do it.


mafintosh:
thats why i get for not writing a test! thanks :)

mafintosh:
released in 3.2.1

Issue #28, opened 2016-02-25
Empty files aren't downloaded


joehand:
Empty files get added to metadata but never get transferred or downloaded.

Reproduce:
- Run `dat link` in dat tests fixtures folder (with empty.txt)
- Download the dat link to another folder.
- folder and empty.txt are missing but metadata gets read (they appear in `createEntryStream`)

mafintosh:
we need to create an empty file here instead of returning right away to fix this, https://github.com/mafintosh/hyperdrive/blob/master/index.js#L159

joehand:
Fixed in 3.3.1

joehand:
Same issue as before. They are added to metadata but never get created.

mafintosh:
We should add a test for this then

clkao:
It seems [archive.download](https://github.com/mafintosh/hyperdrive/blob/master/archive.js#L352) needs to special case where `start == end`.  I tried simply doing:
```
        self.options.storage.openAppend(entry.name, true)
        self.options.storage.closeAppend(cb)
```

but that doesn't seem to work as raf thinks it's noop.

PR #30, opened 2016-02-26
Copy empty folders and files (#28)

joehand:
Fixes empty file and directory bug (#28)

I added one test for the empty file. I was having trouble testing the empty directory issue though - can you run `appendFile` on a directory somehow?

### DIFF
--- a/index.js
+++ b/index.js
@@ -161,7 +161,9 @@ Archive.prototype.download = function (i, cb) {
     if (!feed) return createEmtpyEntry()
 
     function createEmtpyEntry () {
-      mkdirp(path.dirname(dest), function (err) {
+      var dir = dest
+      if (entry.type === 'file') dir = path.dirname(dest)
+      mkdirp(dir, function (err) {
         if (err) return cb(err)
         if (entry.type !== 'file') return cb(null)
         fs.open(dest, 'a', function (err) {
--- a/index.js
+++ b/index.js
@@ -158,20 +158,7 @@ Archive.prototype.download = function (i, cb) {
     var feed = self._getFeed(entry)
     var dest = join(self.directory, entry.name)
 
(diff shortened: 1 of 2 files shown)


mafintosh:
Yea appendFile works with directories as well (which makes the name a bit silly)

index.js
```diff
@@ -156,7 +156,22 @@ Archive.prototype.download = function (i, cb) {
...
+
+    if (!feed) return createEmtpyEntry()
+
+    function createEmtpyEntry () {
+      var dir = dest
+      if (entry.type === 'file') dir = path.dirname(dest)
+      mkdirp(dir, function (err) {
+        if (err) return cb(err)
+        if (entry.type !== 'file') return cb(null)
+        fs.open(dest, 'a', function (err) {
```
mafintosh: fs.open calls the callback with a file descriptor as well. you should close that afterwards

```
fs.open(..., function (err, fd) {
  if (err) return cb(err)
  fs.close(fd, cb)
})
```

index.js
```diff
@@ -156,7 +156,22 @@ Archive.prototype.download = function (i, cb) {
     if (err) return cb(err)
 
     var feed = self._getFeed(entry)
-    if (!feed) return cb(null)
+    var dest = join(self.directory, entry.name)
+
+    if (!feed) return createEmtpyEntry()
+
+    function createEmtpyEntry () {
```
mafintosh: small nit pick. could you move this to the btm of the function with the other funcs?

mafintosh: also i think you typo'ed the name hehe

mafintosh:
Looking really good!

index.js
joehand: Sure, will move down. haha thanks for catching that.

mafintosh:
3.3.1

PR #31, opened 2016-02-27
Add ability to lookup a file by name in an archive.

karissa:


index.js
```diff
@@ -135,6 +135,22 @@ Archive.prototype.close = function (cb) {
...
 }
 
+Archive.prototype.lookup = function (fileName, cb) {
+  // TODO: use binary search or something more performant
+  var stream = this.createEntryStream()
+  var _found = false
+  stream.on('data', function (entry) {
+    if (entry.name === fileName) {
+      cb(null, entry)
+      _found = true
```
mafintosh: just do stream.destroy() here

karissa: kk

PR #40, opened 2016-03-11
Normalize windows filenames upon append

karissa:
Borrowed this code from tar-fs, should we put normalize.js into a different module? Ideas on a name?

### DIFF
--- a/index.js
+++ b/index.js
@@ -11,6 +11,7 @@ var through = require('through2')
 var pump = require('pump')
 var pumpify = require('pumpify')
 var util = require('util')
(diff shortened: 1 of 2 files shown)


mafintosh:
there is already a [normalize-path](https://github.com/jonschlinkert/normalize-path) module that does this i discovered yesterday.
wanna just update the pr to use that?

karissa:
OK, I changed it to use normalize-path, although we still don't have windows CI running.

mafintosh:
awesome. released in 4.1.0

PR #50, opened 2016-05-20
expose the hyperdrive archive object in the external api

karissa:
this allows me to do something like

```
var archive = require('hyperdrive/archive') 
```

if I just want to get at the archive object.


mafintosh:
what's the use case for using the archive prototype directly? it still needs the drive passed into the constructor

karissa:
I had the idea of overwriting one method or adding a method to the archive object, but couldn't get to it.

juliangruber:
it's more tidied up to have `Archive` in it's own file, no matter the use case, so +1 from me

about the use case, is that for experimentation? otherwise it would show an api sufficiency that really we should cover.

index.js
```diff
@@ -1,15 +1,5 @@
...
 var hypercore = require('hypercore')
-var bulk = require('bulk-write-stream')
-var from = require('from2')
-var rabin = process.browser ? require('through2') : require('rabin')
-var pump = require('pump')
-var pumpify = require('pumpify')
-var collect = require('stream-collector')
-var messages = require('./messages')
-var storage = require('./storage')
+var Archive = require('./archive.js')
```
mafintosh: could you remove the .js so it fits with the other requires?

mafintosh:
yep i'm fine splitting it into a separate file but beware that no guarantees about the filename/internal/undocumented methods are given and might be break at any time.

@karissa i've commented a small nitpick, otherwise ltgm

karissa:
ok done, thank you

karissa:
as for the use case, it was for experimentation in github.com/karissa/hypercommit, but I could see it being useful if I wanted to add extra features to a hyperdrive archive. What do you think about that use?

mafintosh:
i wouldn't add any methods to the archive outside of this module in general. your use case makes sense but not sure how i would impl that yet

PR #51, opened 2016-05-27
list() offset

substack:
This patch allows `archive.list()` to accept an `opts.offset` parameter instead of the default value of `0`. This way, I can more easily write a hyperdrive indexer that tails the internal feed without having to implement the machinery already defined in `.list()`.


juliangruber:
+1

substack:
https://github.com/substack/hyperdrive-index

mafintosh:
Great feature. 5.2.0

PR #59, opened 2016-06-07
Adding entry.content with bytesOffset and blockOffet

andrewosh:
When a new entry is added to an archive, the `content` field of the entry message is populated with that entry's `bytesOffset` and `blockOffset` -- just pulled from the the hypercore on insertion. 

Added a few tests for archives containing one and two files in `test/misc.js`.


karissa:
nice

PR #66, opened 2016-06-11
make createByteCursor synchronous

andrewosh:
This PR changes createByteCursor into a synchronous call that will asynchronously get an entry out of the archive the first time `next()` is called.

@mafintosh this look better? It's certainly been easier to work with so far.


mafintosh:
this is great. i'm gonna change the offset to be an options map because i wanna add a few more things

PR #67, opened 2016-06-12
Make example in README.md easier to follow

poga:
Hi, first time hyperdrive user here.

Based on my experience trying hyperdrive, here's my attempt to make the example in `README.md` easier to follow.

1. The `'another-hyperdrive.db'` example led me to think it's in another file. However, it should be put right after the `archive.finalize` part, so I merged them together.
2. Copy-paste and you're ready to go. Better experience for anyone just trying out.

Any comment is appreciated :)

### DIFF
--- a/README.md
+++ b/README.md
@@ -12,11 +12,12 @@ If you are interested in learning how hyperdrive works on a technical level a sp
 
 ## Usage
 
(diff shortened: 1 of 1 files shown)


mafintosh:
looking great!

PR #74, opened 2016-07-01
checkout

substack:
This patch implements `archive.checkout(hashOrBlock)`, which returns a read-only sub-archive rooted at a rolling root hash or block.

depends on https://github.com/mafintosh/hypercore/pull/30

### DIFF
--- a/test/checkout.js
+++ b/test/checkout.js
@@ -1,0 +1,55 @@
+var test = require('tape')
+var hyperdrive = require('../')
+var memdb = require('memdb')
+var concat = require('concat-stream')
+
+test('checkout', function (t) {
+  var contents = [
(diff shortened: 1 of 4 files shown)


test/checkout.js
```diff
@@ -0,0 +1,55 @@
...
+    if (--pending === 0) done()
+
+    function done () {
+      archive.metadata.head(function (err, hash) {
+        t.error(err)
+        hashes.push(hash)
+        next(i + 1)
+      })
+    }
+  })(0)
```
pguth: Ha, funny 😸

substack:
implemented in userland instead: https://github.com/substack/hyperdrive-checkout

Issue #53, opened 2016-06-04
Useful events


somebody1234:
- [ ] When an archive is finished downloading
 - [ ] When an update arrives
 - [ ] When an update is finished downloading (can be the same event as when the archive is finished downloading)

mafintosh:
We have the first event now (`metadata.on('append', ...`), and will have the other one soon

PR #78, opened 2016-07-11
Add file and download-finished events (#53)

joehand:
### New Events:
- `file-downloaded`
- `file-appended`
- `download-finished`
- `finalized`

Plus docs and event tests.

@mafintosh before you suggested: 

>  good pr would be to add file-download, file-upload, file-downloaded events

What's the difference b/t file-download and file-downloaded? And what would someone use file-upload event for (or did you mean file-appended)?

### DIFF
--- a/README.md
+++ b/README.md
@@ -163,6 +163,18 @@ Emitted every time a piece of data is downloaded
 
 Emitted every time a piece of data is uploaded
 
+#### `archive.on('file-download-finished', entry)`
+
+Emitted every time an entry is downloaded
+
+#### `archive.on('download-finished')`
+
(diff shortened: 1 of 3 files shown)


mafintosh:
@joehand `file-download` was when a chunk was downloaded for a file, but it seems we dont need that anymore with the new ui. `file-uploaded` was when a file was fully uploaded to another peer (we can add that later)

clkao:
The description on `download-finished` for live is unclear about what 'updates finish' means.  Is it per metadata-feed change that finishes its content update?

karissa:
what's the status on this one? are we able to resolve the conflicts?

joehand:
This is pretty stale. I think we added similar events but can revisit if we need these again.

Issue #77, opened 2016-07-08
Document archive.live


pfraze:
(According to jhand) you can see if an archive is live with the `.live` attribute. This isn't in the docs, atm.

mafintosh:
This is indeed true. Note that its only populated after archive.open(cb) has been fired. A PR that adds this to the readme would be appreciated.

mafintosh:
fixed in master

Issue #72, opened 2016-06-29
archives should be live per default


mafintosh:

mafintosh:
fixed in master

PR #79, opened 2016-07-11
make live default and add live docs (#72, #77)

joehand:
* Live by default
* document `archive.live`

### DIFF
--- a/archive.js
+++ b/archive.js
@@ -27,7 +27,7 @@ function Archive (drive, key, opts) {
 
   this.options = opts || {}
   this.drive = drive
-  this.live = this.options.live = !!this.options.live
+  this.live = this.options.live = (this.options.live !== false)
   this.metadata = drive.core.createFeed(key, this.options)
   this.content = null
   this.key = key || this.metadata.key
--- a/test/overwrite.js
+++ b/test/overwrite.js
@@ -6,7 +6,7 @@ var concat = require('concat-stream')
 test('overwrite', function (t) {
   t.plan(2)
   var drive = hyperdrive(memdb())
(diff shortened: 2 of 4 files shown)


juliangruber:
it would be cool though if there was a test where the `live:true` can now be removed

also travis fails in cases where `opts` isn't defined

juliangruber:
otherwise +1!

karissa:
great!

archive.js
```diff
@@ -27,7 +27,7 @@ function Archive (drive, key, opts) {
 
   this.options = opts || {}
   this.drive = drive
-  this.live = this.options.live = !!this.options.live
+  this.live = this.options.live = opts.hasOwnProperty('live') ? !!this.options.live : true
```
mafintosh: i'd prefer `this.options.live !== false`

joehand:
> it would be cool though if there was a test where the `live:true` can now be removed

@juliangruber do you mean removing `live:true` from existing test? I could do that for [this one](https://github.com/mafintosh/hyperdrive/blob/master/test/overwrite.js#L9).

juliangruber:
awesome, +1!

test/overwrite.js
```diff
@@ -6,7 +6,7 @@ var concat = require('concat-stream')
 test('overwrite', function (t) {
   t.plan(2)
   var drive = hyperdrive(memdb())
-  var archive = drive.createArchive(undefined, { live: true })
+  var archive = drive.createArchive(undefined)
```
mafintosh: remove undefined as well. was just a spacer for the options map :)

PR #82, opened 2016-07-13
Force writing empty file.

### LINKED ISSUE #28: Empty files aren't downloaded

clkao:
Fixes #28

### DIFF
--- a/archive.js
+++ b/archive.js
@@ -355,6 +355,18 @@ Archive.prototype.download = function (entry, cb) {
   this._range(entry, function (err, start, end) {
     if (err) return cb(err)
 
+    if (start === end && entry.type === 'file') {
+      var storage = self.options.storage
+      if (storage) {
+        storage.openAppend(entry.name, true)
+        storage.write(0, '', function (err) {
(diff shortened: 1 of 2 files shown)


juliangruber:
+1

archive.js
```diff
@@ -355,6 +355,18 @@ Archive.prototype.download = function (entry, cb) {
   this._range(entry, function (err, start, end) {
     if (err) return cb(err)
 
+    if (start === end && entry.type === 'file') {
+      var storage = self.options.storage
+      if (storage) {
+        storage.openAppend(entry.name, true)
+        storage.write(0, '', function (err) {
```
mafintosh: This should be an empty buffer instead of an empty string

mafintosh: fixed in master :D

PR #89, opened 2016-07-21
skip extra entry lookups in _range

substack:
This patch skips the extra recursive lookup in `_range()` for objects which are already entries. This means methods that use `_range` such as `createFileReadStream()` can stream the data from previous documents, not only the latest document.


mafintosh:
awesome stuff, out in 7.1.0

PR #92, opened 2016-07-27
preserve utimes

juliangruber:
i'm totally not sure this is the proper way to implement this...


juliangruber:
@joehand phew! ^ this last commit fixes the mtime behavior in my tests

joehand:
👍   Looks good and is working with hyperdrive-import-files!

mafintosh:
7.3.0

PR #98, opened 2016-09-13
Add archive.countDownloadedBlocks and archive.isEntryDownloaded

### LINKED ISSUE #97: Read file without downloading; or, check if file is downloaded before trying to read
As I understand it, the hyperdrive archive.createFileReadStream will try to download a file before serving it, if it's not already available. Is there a way to disable that, or is there a separate function to check if a file is fully downloaded?

This might be related to the download-progress PRs.

pfrazee:
Closes #97 and adds 2 new methods. From the readme:

---

#### `archive.countDownloadedBlocks(entry)`

Count the number of blocks in the entry that have been downloaded.
You can calculate the file's download progress, as a percentage, with:

```js
var downloaded = archive.countDownloadedBlocks(entry)
var progress =  downloaded / entry.blocks
```

#### `archive.isEntryDownloaded(entry)`

Has all of the entry's blocks been downloaded?

### DIFF
--- a/README.md
+++ b/README.md
@@ -230,6 +230,20 @@ Options include:
 }
 ```
 
+#### `archive.countDownloadedBlocks(entry)`
+
+Count the number of blocks in the entry that have been downloaded.
+You can calculate the file's download progress, as a percentage, with:
+
+```js
+var downloaded = archive.countDownloadedBlocks(entry)
+var progress =  downloaded / entry.blocks
+```
+
+#### `archive.isEntryDownloaded(entry)`
+
+Has all of the entry's blocks been downloaded?
+
 ## License
 
 MIT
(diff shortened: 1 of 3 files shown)


juliangruber:
👍 !

What do you think about making the `entry` object for `countDownloadedBlocks` optional, in which case it would work on the whole archive?

pfrazee:
@juliangruber yeah, that's a good idea, I'll do that

pfrazee:
@juliangruber that change will be easier to implement when the list/history change in https://github.com/mafintosh/hyperdrive/pull/99 is merged. I suggest we merge this as-is, and make an issue to make `entry` optional, to be done after 99.

mafintosh:
7.5.0 - awesome work

PR #102, opened 2016-09-19
Versioning API (WIP)

pfrazee:
This PR is a work-in-progress. It will add version checkpoints, which Hyperdrive's API can create and list, and use to find specific versions of files in its history.

I've written a new message-type in the schema, and updated the readme with the API that I propose for this. I need the changes to `list()` to be merged (https://github.com/mafintosh/hyperdrive/pull/99) before I can finish updating the API (`list` will take a 'version' option).

Let me know if you think anything should change!

### DIFF
--- a/README.md
+++ b/README.md
@@ -128,12 +128,7 @@ Boolean whether archive is live. `true` by default. Note that its only populated
 
 #### `archive.version`
 
-The label of the most recent checkpoint.
-
-#### `archive.semver`
-
-The label of the most recent semantic-version checkpoint.
-This can differ from `archive.version` when the most recent checkpoint was not a semver. 
+The label of the most recent checkpoint. Note that its only populated after archive.open(cb) has been fired.
 
 #### `archive.append(entry, callback)`
 
@@ -156,23 +151,23 @@ archive.append('hello.txt', function () {
 })
 ```
 
-#### `archive.checkpoint(version, [callback])`
+#### `archive.checkpoint(desc, [callback])`
 
-Write a version checkpoint to the archive, marking the state for future reference. Version includes:
+Write a version checkpoint to the archive, marking the state for future reference. Desc includes:
 
 ```js
 {
-  label: '1.0.0', // can be any string, but semantic versions are recommended
+  version: '1.0.0', // can be any string, but semantic versions are recommended
   message: 'The first version!' // an optional string describing the checkpoint
 }
 ```
 
-If the label has been used before, Hyperdrive will error.
-If the label is a semantic version, Hyperdrive will error if it isn't a greater version than all previous semver checkpoints.
+If the version has been used before, Hyperdrive will error.
+If the version is a [Semantic Version](http://semver.org/), Hyperdrive will error if it isn't a greater version than all previous semver checkpoints.
 
 You can mix semvers with non-semver labels.
 
-Hyperdrive automatically adds a `.timestamp` value to `version`, using `Date.now()`.
+Hyperdrive automatically adds a `.timestamp` value to `desc`, using `Date.now()`.
 You can override this by setting your own `.timestamp` if needed.
 
 #### `archive.finalize([callback])`
--- a/README.md
+++ b/README.md
@@ -154,7 +154,7 @@ Write a checkpoint to the archive, marking the state for future reference. Desc 
 ```js
 {
   name: '1.0.0', // can be any string
-  message: 'The first version!' // an optional string describing the checkpoint
+  description: 'The first version!' // an optional string describing the checkpoint
 }
 ```
 
--- a/README.md
+++ b/README.md
@@ -126,6 +126,15 @@ Otherwise it is a 32 byte hash.
 
 Boolean whether archive is live. `true` by default. Note that its only populated after archive.open(cb) has been fired.
 
+#### `archive.version`
+
+The label of the most recent checkpoint.
+
+#### `archive.semver`
+
+The label of the most recent semantic-version checkpoint.
+This can differ from `archive.version` when the most recent checkpoint was not a semver. 
+
 #### `archive.append(entry, callback)`
 
 Append an entry to the archive. Only possible if this is an live archive you originally created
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -117,6 +117,15 @@ Otherwise it is a 32 byte hash.
 
 Boolean whether archive is live. `true` by default. Note that its only populated after archive.open(cb) has been fired.
 
+#### `archive.version`
+
+The label of the most recent checkpoint.
```
juliangruber: maybe document that this is only available after it has finished opening?

README.md
```diff
@@ -117,6 +117,15 @@ Otherwise it is a 32 byte hash.
...
+
+#### `archive.semver`
+
+The label of the most recent semantic-version checkpoint.
```
juliangruber: the semver thing sounds like an addition that could be useful too but might better be discussed in a separate pull request?

README.md
```diff
@@ -138,6 +147,25 @@ archive.append('hello.txt', function () {
...
+Write a version checkpoint to the archive, marking the state for future reference. Version includes:
+
+```js
+{
+  label: '1.0.0', // can be any string, but semantic versions are recommended
+  message: 'The first version!' // an optional string describing the checkpoint
+}
+```
+
+If the label has been used before, Hyperdrive will error.
```
juliangruber: to me `label` isn't a term for an unique datum, maybe let's use `version` instead?

karissa:
What's the difference between version and semver?

karissa:
That was my first question. Maybe having a different name for version would be helpful or removing the "semver" property would be less confusion. Wasn't entirely clear to me at first

README.md
pfrazee: Yep

README.md
pfrazee: I think `version` might be fine in this API call, but I think the message schema should still use `label`. This is the schema:

```
message Checkpoint {
  required string label = 1;
  required uint64 timestamp = 2;
  optional string message = 3;
}
```

`version` wouldn't make sense. It should be `id` or `name`.

pfrazee:
I'll remove `.semver` for now

pfrazee:
Pushed some updates based on the comments

juliangruber:
hm yeah i get your point about using `label`, but then i think we shouldn't call it `version` at all. label is a broader term that works for more use cases.

pfrazee:
That is a possibility: just remove all 'version' semantics entirely from hyperdrive. They're purely checkpoints that add labels to points in the history, and they can be used downstream to implement versions.

In that case, would we have `.checkpoint()` enforce semver monotonicity? We could, I suppose, and still remove most other semantics about versions.

joehand:
> That is a possibility: just remove all 'version' semantics entirely from hyperdrive. They're purely checkpoints that add labels to points in the history, and they can be used downstream to implement versions.

👍  on this. After hearing about versions from RDA, there are 1001 way scientists/publishers/etc. want to do versions. If we allow a checkpoint function and some way to get all checkpoints (so they can enforce whatever downstream) that would be ideal.

joehand:
> That is a possibility: just remove all 'version' semantics entirely from hyperdrive. They're purely checkpoints that add labels to points in the history, and they can be used downstream to implement versions.

👍  on this. After hearing about versions from RDA, there are 1001 way scientists/publishers/etc. want to do versions. If we allow a checkpoint function and some way to get all checkpoints (so they can enforce whatever downstream) that would be ideal.

pfrazee:
Ok that's cool with me. Do we want to consider calling it a "tag" instead
of a "checkpoint," to match the semantics from Git ?

joehand:
Right now we have these ideas:

* Live (latest)
* Snapshot
* Checkpoint (snapshot + label/message)

That seems to make sense to me.

pfrazee:
What's a snapshot?

juliangruber:
@pfrazee like a checkpoint but without a label. just an archive at one point in time



> That is a possibility: just remove all 'version' semantics entirely from hyperdrive. They're purely checkpoints that add labels to points in the history, and they can be used downstream to implement versions.
> 
> In that case, would we have .checkpoint() enforce semver monotonicity? We could, I suppose, and still remove most other semantics about versions.

I don't fully understand this, are you saying to remove versions or to add semver?

karissa:
Let's try to keep it as simple as possible. Enforcing semver monotonicity sounds like a feature that could be nice but some people might want to turn it off. So maybe handle in a separate PR?

Everything looks great, I'd say the names matter less. 

Checkpoint is more like a commit (+metadata).
Snapshot is like a version of the repo, doesn't have metadata. It's just a hash. I'd say its more for internal use or "oopsie" moments where you want to revert. Kinda like google docs or dropbox versions. Let me know if this isn't right @joehand

joehand:
> Checkpoint is more like a commit (+metadata).
Snapshot is like a version of the repo, doesn't have metadata. It's just a hash. I'd say its more for internal use or "oopsie" moments where you want to revert. Kinda like google docs or dropbox versions. Let me know if this isn't right @joehand

Exactly!

pfrazee:
The version semantics are removed

karissa:
nice! ltm

schema.proto
```diff
@@ -34,3 +34,9 @@ message Entry {
 //}
 
 // missing: truncate, chown, chmod,
+
+message Checkpoint {
+  required string name = 1;
+  required uint64 timestamp = 2;
+  optional string message = 3;
```
juliangruber: what about `name`, `timestamp` and `description`?

pfrazee: Sure

karissa:
this + history api will make me so happy

karissa:
itching to merge

pfrazee:
@karissa :) I was waiting for #99 to merge before writing this PR's code, but I'll go ahead and get that done now

pfrazee:
Checkpoint() and checkpoints() are now implemented. DONT MERGE BEFORE #99 IS MERGED! There are new messages, with new numbers, assigned in both PRs. (Unlink in the other, Checkpoint in this.) Currently they both have the same message number, and that no good.

pfrazee:
Todo: update checkpoint msg to hyperdrive-encoding

pfrazee:
Todo: add a hash of the content using https://github.com/mafintosh/hypercore/blob/master/lib/feed.js#L152

pfrazee:
I rebased this PR to latest on master. Here's the message schema I used, which will need to be moved into hyperdrive-encodings:

```
message Checkpoint {
  required string name = 1;
  required uint64 timestamp = 2;
  optional string description = 3;
} 
```

mafintosh:
See .checkout in 8

PR #104, opened 2016-09-29
refactor to use hyperdrive-encoding

juliangruber:


mafintosh:
👍 we should change the interface of the encoding methods to be more clear now that it is a public module but good enough for now

mafintosh:
#letsmajorbumpthatlater

juliangruber:
> 👍 we should change the interface of the encoding methods to be more clear now that it is a public module but good enough for now

can you explain in https://github.com/juliangruber/hyperdrive-encoding/issues/new?

PR #106, opened 2016-09-30
upgrade hyperdrive-encoding

juliangruber:


archive.js
```diff
@@ -80,7 +80,7 @@ Archive.prototype.list = function (opts, cb) {
       if (err || !buf) return cb(err, buf)
 
       try {
-        var entry = encoding.decodeEntry(buf)
+        var entry = encoding.decode(buf)
```
juliangruber: no `entry.type` checks necessary for those as before it also just checked whether it was _any_ valid message.

PR #111, opened 2016-10-18
Fix doc about default value of live of archive.list()

poga:
```
  var live = opts.live === false ? false : (opts.live || !cb)
```

archive.list() have different default for `live` option based on callback is given or not
- `archive.list(cb)` : `live` is false
- `var rs = archive.list()` : `live` is true


pfrazee:
Yeah this is a weird behavior, IMO. I think it'd be better to have live always default to false, yeah?

poga:
Yeah, I agree.

I just want to document the behavior before we decided how it should work (which might break some dependents)

pfrazee:
Definitely, thanks for doing that

PR #116, opened 2016-11-30
allow feeds to be passed to createArchive directly

joehand:
Hypercore feeds can be passed to `createArchive` as `metadata` and `content` options.

The test is pretty basic. Let me know if there is other things that would be good to test here.


mafintosh:
7.11.0

PR #122, opened 2017-01-12
Add createFileWriteStream opts & indexing docs

joehand:


karissa:
+1

pfrazee:
Actually could you explain a bit more in the docs there? Im still not sure what it does.

karissa:
Yeah, it is a bit confusing how it's written.

joehand:
How about:

> If the file source and archive destination are the same, it is a good idea to set `indexing: true`. This will speed up the write stream by not writing the file to the file system. With `indexing: false`, the file is written to the file system while it's imported to the archive (useful if copying the file from a separate source).

pfrazee:
Oh ok I see what that is. It'd be helpful if the flag was named `indexOnly`.

Riffing off of yours:

> If indexing: true, the write-stream will add the file metadata to the archive, but will not write the actual file contents. This is useful when creating an archive "in-place." It should be set to false if replicating the archive to a new destination.

karissa:
It's not just just a new destination of the archive, it's useful at the file level if the file already exists on the file system. 8 like joe's

PR #123, opened 2017-01-19
pass metadata block seq to .file

mafintosh:
adds `file(name, archiveOptions, seq)`

### DIFF
--- a/archive.js
+++ b/archive.js
@@ -389,7 +389,7 @@ Archive.prototype.append = function (entry, cb) {
     if (entry.type === 'file') {
       if (!self.options.storage) throw new Error('Set options.file to append files')
 
-      var rs = fileReadStream(self.options.file(entry.name, self.options))
+      var rs = fileReadStream(self.options.file(entry.name, self.options, self.metadata.blocks))
       var ws = self.createFileWriteStream(entry, {indexing: true})
       pump(rs, ws, cb)
     } else {
--- a/storage.js
+++ b/storage.js
@@ -74,7 +74,7 @@ Storage.prototype._open = function (offset, length, buf, cb) {
   var result = {
     start: offset,
     end: Infinity,
-    storage: this.archive.options.file(this._appendingName, this.archive.options)
+    storage: this.archive.options.file(this._appendingName, this.archive.options, this.archive.metadata.blocks)
   }
   this._appendingName = null
   this._appending = result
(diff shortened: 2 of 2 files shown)


pfrazee:
It's not wholly impossible that `seq` will get reused if you have multiple concurrent writes, is it?

mafintosh:
no, it should always be unique

mafintosh:
@pfrazee one thing to consider is adding a flag of some sorts that tells you if this is the latest version (i.e. if you are reading a file fx) but maybe we can add that later

pfrazee:
Are you sure? What if I did:

```
archive.append(...)
archive.append(...)
```

The files have to be processed before the metadata is updated, dont they?

mafintosh:
i meant in theory (that should run inside a mutex). there might be bugs!

pfrazee:
Is this the only thing that would require us to write in a mutex? Or did we already need to use a mutex?

mafintosh:
we already need to

pfrazee:
Wait really? Does that need to be done by the API consumer, aka do I need to add mutexes to beaker?

mafintosh:
No, there is an internal one. The write streams / .append will simply pause while a single one is being processed

pfrazee:
Oh I see. Ok. Let me review again...

pfrazee:
Ok how would this work on read() ? Does bisect() get called during open if you're reading?

mafintosh:
same as it works now. bisect is called if a block is being replicated and you don't know which file it belongs to

pfrazee:
What I'm wondering is, how do we control which version is being read? It looks like metadata.blocks is always being used, which works for writes but not for reads -- except in bisect.

(Not at all to be critical: Im having some trouble groking this code, I think it could use a refactor.)

mafintosh:
metadata.blocks is only used for append/writes (it's the latest version - metadata will be stored at that seq). seq just always points to the metadata index the current version is being stored at

mafintosh:
you would just write data to `filename.{seq}` to have unique filenames per version

pfrazee:
> seq just always points to the metadata index the current version is being stored at

That doesnt seem to be the case though. What's getting passed into file() as seq is the `metadata.blocks`, which makes sense for the write, but for the read it ought to be the sequence-number of the metadata-entry for that file, right?

pfrazee:
How does createFileReadStream() correctly identify the file to read?

mafintosh:
the bisect indirectly makes that happen. the metadata.blocks is only passed to writes

mafintosh:
we can pass something else there if you think it is confusing. it just indicates the version it will be stored at.

pfrazee:
Yeah Im just trying to make sure it's correct. So, when createFileReadStream happens, bisect() is used... wait is bisect a search for the entry? Oh man, it is isnt it

pfrazee:
Ok that looks like it should work. We'll need a way to specify the revision later.

pfrazee:
We should add some tests to the PR.

PR #124, opened 2017-02-09
add replicate test with file updates

joehand:
Adds two tests, one of them is failing (the one that uses `raf` options for the length stuff).


joehand:
Also, worth noting, if you use `archive.download(0)` and then `archive.createReadStream` directly then all is okay. Its only when I waited for replication to finish automatically that it didn't work.

PR #127, opened 2017-02-17
stop selectLatest if archive is closed

joehand:
`_selectLatest` keeps trying to download stuff even after the archive is closed. This is causing lots of seg faults when we try to close things. 

PR makes sure we stop loops if archive is closed.


mafintosh:
Fixed a minor bug in this PR and it is out in 7.14.3

PR #128, opened 2017-02-21
Fix sparse archive option with file option

joehand:
Right now `sparse` doesn't work if the `file` option is also set because `_selectLatest` gets called. PR adds a test and fix.

* Adds a test using both `opts.sparse: true` and `opts.file`
* Set `_onlyLatest` to false if sparse is true


juliangruber:
@joehand you rock!

pfrazee:
Just tested this in Beaker and it works. Definitely need it, can we merge?

mafintosh:
7.14.5

PR #129, opened 2017-02-23
Normalize paths on lookup()

pfrazee:
The hyperdrive ecosystem doesnt enforce whether there should be a preceding slash on file entries in the metadata stream. This means that lookups can sometimes fail because of different policies on preceding slashes. This PR fixes that in lookup().

(Solves a problem in bkr)


mafintosh:
That was a bug! 7.14.4

PR #132, opened 2017-04-06
add storage tests for secret key bug

joehand:
* Add a [failing test](https://github.com/mafintosh/hyperdrive/compare/master...joehand:secret-key-test?expand=1#diff-6a1112979275ee9654c3b1f5a4c15c2dR16) for secret key bug
* Also adds a few basic storage tests

Not sure how best to *resume* an existing archive with v8 API, but this seems to work =). Passing secret key in https://github.com/mafintosh/hyperdrive/blob/master/index.js#L393-L395 fixes bug + passes test.

### DIFF
--- a/package.json
+++ b/package.json
@@ -19,7 +19,8 @@
   "devDependencies": {
     "random-access-memory": "^2.3.0",
     "standard": "^9.0.2",
-    "tape": "^4.6.3"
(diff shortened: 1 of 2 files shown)


mafintosh:
Fixed in 8.0.1 with the method you describe

PR #139, opened 2017-04-10
Add archive.writable attribute

pfrazee:


mafintosh:
can we call this writable instead? then its closer to hypercore

pfrazee:
Yeah, done & pushed

PR #141, opened 2017-04-10
Readme: Add origKey to replicate example

pfrazee:
Based on some errors I just got, it looks like including the original archive's key is required for replication


mafintosh:
Ah right

PR #150, opened 2017-04-15
docs: add archive.ready(cb)

juliangruber:


ralphtheninja:
I know this is technically part of the API since `.ready` is not a "private" property. Is it _meant_ to be part of the api or should we hide it? I think it's a bit confusing that it's called potentially with an error. We already have events for `'ready'` and `'error'` which should be enough.

ralphtheninja:
@juliangruber `.ready` was deprecated in favor of the `'ready'` event, so closing this one.

PR #164, opened 2017-05-16
add test for using sparse and createReadStream

karissa:
add tests for sparse


ralphtheninja:
@karissa I rebased it onto latest master.

LGTM!

PR #169, opened 2017-05-29
Support .open, .read and a high level file cursor api

mafintosh:
WIP, opened now so people can follow along and give feedback


mafintosh:
9.7.0

PR #173, opened 2017-06-24
Fix exception-throwing edgecase in createReadStream

pfrazee:
Closes https://github.com/mafintosh/hyperdrive/issues/172


mafintosh:
Note to self that this should be fixed in hypercore as well (cb should be a noop)

mafintosh:
9.4.6

PR #176, opened 2017-07-26
Set key when passing metadata, content feeds directly

### LINKED ISSUE #175: archive.key does not get set if you pass in content
If you pass in `content` as an option, the archive.key [never gets set](https://github.com/mafintosh/hyperdrive/blob/master/index.js#L696). May also be fixed with #149. 

I can PR test for this. Seems like it'd be an easy fix moving the self.key above cb in line linked, but not sure if there is another constraint I'm missing.

joehand:
Creating archive from feeds, key is never set:

```js
var archive = hyperdrive(storage, {metadata: source.metadata, content: source.content})
archive.key // null
archive.ready(function () {
  archive.key // null
})
```

Fixes #175

### DIFF
--- a/index.js
+++ b/index.js
@@ -693,10 +693,10 @@ Hyperdrive.prototype._open = function (cb) {
     if (err) return cb(err)
     self.metadata.ready(function (err) {
       if (err) return cb(err)
-      if (self.content) return cb(null)
-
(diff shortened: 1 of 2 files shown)


andrewosh:

PR #180, opened 2017-08-09
Pass sparse flag to metadata as well as content hypercore

creationix:
When replicating to a browser via websockets and storing in indexedDB, performance isn't the best.  It would be nice if the metadata could be sparsely replicated in addition to the content.


ralphtheninja:
lgtm

mafintosh:
9.5.0

PR #185, opened 2017-08-17
remove legacy version setter

### LINKED ISSUE #183: Setting version with getter in update
The update method for hyperdrive looks like the following
```
  function update () {
    self.version = self.tree.version
    self.emit('update')
  }
```
The version property only has a getter so this seems a bug.

scriptjs:
Tests remain unchanged and pass. Fixes #183


Armen0705:
scriptjs:master

PR #186, opened 2017-08-17
(WIP) add tests using feed.default to switch indexing

### LINKED ISSUE #160: indexing should be option on write stream, not hyperdrive
`opts.indexing` doesn't make sense on the wholee hyperdrive. We may want to import in place and from another directory for the same archive. It should be on the write stream instead.

joehand:
Adds tests for hypercore indexing switching.

Fixes #160 (kind of). Should I add this as an option to `archive.createWriteStream` also? We could do that by setting the `archive.content.default(writeOpts)` then resetting them.

**WIP**: Requires https://github.com/mafintosh/hypercore/pull/115 to be released

### DIFF
--- a/package.json
+++ b/package.json
@@ -20,6 +20,7 @@
     "unixify": "^1.0.0"
   },
   "devDependencies": {
+    "dat-storage": "^1.0.3",
     "random-access-memory": "^2.3.0",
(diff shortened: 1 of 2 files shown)


joehand:
This will change with hyperdb, closing.

PR #192, opened 2017-10-01
Expose a download() method on the archive to download all contents in a version

e-e-e:


test/basic.js
```diff
@@ -116,3 +116,43 @@ tape('provide keypair', function (t) {
     })
   })
 })
+
+tape.only('download a version', function (t) {
```
mafintosh: can you remove the .only here?

index.js
```diff
@@ -301,6 +301,49 @@ Hyperdrive.prototype.checkout = function (version) {
...
+Hyperdrive.prototype.download = function (cb) {
+  var downloadCount = 1
+  var self = this
+
+  download('/')
+
+  function download (entry) {
+    self.stat(entry, function (err, stat) {
+      if (err) {
+        self.emit('error', err)
```
mafintosh: Don't emit an error, just call the cb

index.js
```diff
@@ -301,6 +301,49 @@ Hyperdrive.prototype.checkout = function (version) {
...
+      }
+      if (stat.isDirectory()) return downloadDir(entry, stat)
+      if (stat.isFile()) return downloadFile(entry, stat)
+    })
+  }
+
+  function downloadDir (dirname, stat) {
+    self.readdir(dirname, function (err, entries) {
+      if (err) {
+        self.emit('error', err)
```
mafintosh: Same as above

index.js
e-e-e: ok, easy.

mafintosh:
9.9.0

PR #193, opened 2017-10-05
add failing sparse test

joehand:


mafintosh:
should be fixed when merged (a dependency was the issue)

PR #201, opened 2018-01-23
Immediately register content hypercore error event

pfrazee:
If registering for 'error' is delayed then in some cases (such as the "Key and secret do not match") the listener is registered too late and the app doesn't get a chance to handle it (it throws toplevel).

This PR makes sure that hyperdrive catches and bubbles the event.


mafintosh:
9.12.2

PR #204, opened 2018-03-18
Multiwriter Hyperdrive (hyperdb backend)

mafintosh:
Still missing APIs but close to all functionality is here now. Added a bunch more tests as well.

Multiwriter is provided by switching to [hyperdb](https://github.com/mafintosh/hyperdb) as the backend instead of append-tree. This also fixes most of our known scaling issues in term of storing lots of files in a single directory and more.

Prob the main thing missing is testing that our files-as-files storage still work and "latest" file tracking (mainly just porting this over from master)

/cc @joehand


mafintosh:
Oh and docs for multiwriter are missing hehehe.

Mainly involves calliing `archive.authorize(otherKey)`

jimpick:
I see:

```js
Hyperdrive.prototype.authenticate = function (key, cb) {
  this.db.authenticate(key, cb)
}
```

But hyperdb has no .authenticate() method. Was this intended to be .authorize() ?

andrewosh:

PR #206, opened 2018-03-22
Replace buffer constructor with buffer-alloc and buffer-from

jasnell:


juliangruber:
has anyone thought about writing a migration script 😁

jasnell:
Yes but to be honest I don't think it would really save that much effort. Each of these needs to be looked at individually to determine if the `new Buffer()` should become a `Buffer.alloc()`or `Buffer.from()` and we'd have to check the results of the migration script anyway. It likely wouldn't save much time.

andrewosh:

PR #213, opened 2018-04-21
Pick a version when there are conflicts

jimpick:
The existing code doesn't return a value when there is more than one
choice.

This fixes the issue where data would 'disappear' on ttps://dat-multiwriter-web.glitch.me/ when two writers make the same change at the same time.


mafintosh:
wow! embarrassing :grin:

PR #217, opened 2018-05-18
Add cached and wait opts to stat, lstat, access, and exists

pfrazee:
Makes it possible to

```js
archive.stat('/', {wait: true}, cb)
```

and/or

```js
archive.stat('/', {cached: true}, cb)
```

They actually have different effects. `cached=true` seems to succeed if the data isnt available if the block represents a folder, while `wait=false` will error no matter what.

### DIFF
--- a/README.md
+++ b/README.md
@@ -165,7 +165,7 @@ Options include:
 ```js
 {
   encoding: string
-  cached: true|false // default: false,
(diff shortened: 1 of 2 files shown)


mafintosh:
:+1:

pfrazee:
Published `9.13.0`

PR #229, opened 2018-10-04
_latestStorage now closed with archive.

mistersdevelopment:
_latestStorage was not being closed when the archive was closed. This lead to file descriptors being left open. This prevented an archive from being created, closed, and then deleted from disk in the same process.


andrewosh:
Closing because this is outdated (Hyperdrive 9).

PR #231, opened 2018-12-13
Preventing unexpected close error that occurs with newer version of hypercore

martinheidegger:
Hypercore [introduced](https://github.com/mafintosh/hypercore/commit/3ec81f191fd17e89f7ec2884a3eb6cf77c9ac00c) a new error to callbacks when closing.
This should have probably been a breaking change, as it triggers a new error on close in hyperdrive.

This PR adds a test for the new error case 33e0acab6ccee88c5ee77a0ef24f30c0ea94c400 and fixes it by using a flag if the hypercore is closed 79d10ad777a2aeade0af4e7a22b7d3fc8017e78f.


mafintosh:
Thanks! 9.14.1

PR #233, opened 2019-02-06
[Release candidate] v10 prerelease

mafintosh:
@andrewosh and I (but mostly @andrewosh) have been busy getting Hyperdrive v10 ready to ship. It's a major change that moves the internal filesystem structure to [hypertrie](https://github.com/mafintosh/hypertrie), for massive perf boosts.

The is the first "official" step towards multiwriter as the trie enables the data structures needed for that, but v10 will still be a single writer only hyperdrive, focused on getting the trie out so we can get some milage on that.

Also massively increases test coverage.

NOTE: we are still finishing up the last parts of the API

### Prerelease Checklist
- [x] Support for writable file descriptors
- [x] Stateful file descriptor fuzz test(s)
- [ ] Benchmark + test completeness with a barebones FUSE implemention (minus random-access writes)
- [ ] Support for "live" files (writing a stat object at the beginning of the file write with a `live` flag)


index.js
```diff
@@ -1,924 +1,578 @@
...
-
-function wrap (self, storage) {
-  return {
-    metadata: function (name, opts) {
-      return storage.metadata(name, opts, self)
-    },
-    content: function (name, opts) {
-      return storage.content(name, opts, self)
+  checkout (version, opts) {
+    const versionedTrie = this._db.checkout(version)
```
martinheidegger: Why not call it `db` and use it directly in the options.

index.js
```diff
@@ -1,924 +1,543 @@
...
+    this._storages = defaultStorage(this, storage, opts)
+
+    this.metadataFeed = opts.metadataFeed || hypercore(this._storages.metadata, key, {
+      secretKey: opts.secretKey,
+      sparse: !!opts.sparseMetadata,
+      createIfMissing: opts.createIfMissing,
+      storageCacheSize: opts.metadataStorageCacheSize,
+      valueEncoding: 'binary'
+    })
+    this._db = opts.db
```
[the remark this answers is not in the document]
mafintosh: It's not an offical option, used for checkouts internally.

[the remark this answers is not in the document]
martinheidegger: Shouldn't it be `opts._db` then?

index.js
andrewosh: Yep, was a leftover from debugging, but now it's removed. Thanks

lib/content.js
```diff
@@ -0,0 +1,33 @@
+const sodium = require('sodium-universal')
+
+function contentKeyPair (secretKey) {
```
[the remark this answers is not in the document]
andrewosh: Sure is! Removed the `lib/keys` file.

index.js
```diff
@@ -204,6 +206,35 @@ class Hyperdrive extends EventEmitter {
...
+
+    this.contentReady(err => {
+      if (err) return cb(err)
+
+      this._db.get(name, (err, st) => {
+        if (err) return cb(err)
+        if (!st) return cb(new errors.FileNotFound(name))
+
+        // 20 is arbitrary, just to make the fds > stdio etc
+        const fd = 20 + this._fds.push(new FD(this.contentFeed, name, st.value)) - 1
```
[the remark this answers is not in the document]
andrewosh: :+1:

index.js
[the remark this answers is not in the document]
andrewosh: `opts._db` is definitely more suitable, since we aren't exposing it in the API.

yoshuawuyts:
@mafintosh are you planning to release a blog post when v10 is released? Even if it's short, this seems like it might be a good opportunity to do a bit of marketing, and helping others catch up on what's changed.

README.md
```diff
@@ -295,25 +273,12 @@ Similar to fs.access.
...
-#### `archive.open(name, flags, [mode], callback)`
-
-Open a file and get a file descriptor back. Similar to fs.open.
-
-Note that currently only read mode is supported in this API.
-
-#### `archive.read(fd, buf, offset, len, position, callback)`
-
-Read from a file descriptor into a buffer. Similar to fs.read.
-
```
mafintosh: @andrewosh these are back in right?

lib/fd.js
```diff
@@ -0,0 +1,309 @@
...
+      return cb(new errors.BadFileDescriptor('File descriptor not open in create mode.'))
+    }
+    if (this.position !== null && pos !== this.position) {
+      return cb(new errors.BadFileDescriptor('Random-access writes are not currently supported.'))
+    }
+    if (this.appending && pos < this.stat.size) {
+      return cb(new errors.BadFileDescriptor('Position cannot be less than the file size when appending.'))
+    }
+    if (!this._writeStream && !this.appending && pos) {
+      return cb(new errors.BadFileDescriptor('Random-access writes are not currently supported.'))
```
mafintosh: for all these sync calls to the cb above, next tick them, `process.nextTick(cb, new errors...)`

lib/fd.js
```diff
@@ -0,0 +1,309 @@
...
+    }
+
+    const self = this
+    if (!this._writeStream) {
+      const writeStream = this.drive.createWriteStream(this.path)
+      this._writeStream = pumpify(this._batcher, through.obj((chunk, enc, cb) => {
+        cb(null, Buffer.concat(chunk))
```
mafintosh: why is the concat needed here? or is that the implicit copy we talked about?

mafintosh: (also move this into a static function as it has no deps on anything in the scope)

lib/fd.js
```diff
@@ -0,0 +1,309 @@
...
+      return cb(new errors.BadFileDescriptor('Random-access writes are not currently supported.'))
+    }
+
+    const self = this
+    if (!this._writeStream) {
+      const writeStream = this.drive.createWriteStream(this.path)
+      this._writeStream = pumpify(this._batcher, through.obj((chunk, enc, cb) => {
+        cb(null, Buffer.concat(chunk)) 
+      }), writeStream)
+      this._writeStream.on('error', err => {
```
mafintosh: can be 'error', cb no?

lib/fd.js
```diff
@@ -0,0 +1,309 @@
...
+      const slice = buffer.slice(offset, len)
+      write(self._writeStream, slice, err => {
+        if (err) return cb(err)
+        self.position += slice.length
+        return cb(null, slice.length, buffer)
+      })
+    }
+
+    function write(stream, data, cb) {
+      if (!stream.write(data)) {
```
mafintosh: to be super streamy you actually have to do stream.write(data) === false for this check

lib/fd.js
```diff
@@ -0,0 +1,309 @@
...
+    process.nextTick(cb, null)
+  }
+
+  /**
+   * Will currently request the next 16 blocks linearly.
+   *
+   * TODO: This behavior should be more customizable in the future.
+   */
+  _refreshDownload (start, cb) {
+    let end = Math.min(this.stat.blocks + this.stat.offset, start + 16)
```
mafintosh: const

lib/fd.js
```diff
@@ -0,0 +1,309 @@
...
+    }
+
+    this._range = this.drive.contentFeed.download({ start, end, linear: true }, cb || noop)
+  }
+
+  _seekAndRead (buffer, offset, len, pos, cb) {
+    const start = this.stat.offset
+    const end = start + this.stat.blocks
+
+
```
mafintosh: too much whitespace (i think standard should catch this?)

lib/fd.js
```diff
@@ -0,0 +1,309 @@
...
+          this.blockOffset = blkOffset + read
+        }
+      }
+
+      cb(null, read, buffer)
+    })
+  }
+}
+
+FileDescriptor.create = function (drive, name, flags, cb) {
```
mafintosh: If I'm not mistaken this is the only func we actually use outside this module? If so we should module.exports it instead of the class.

lib/fd.js
```diff
@@ -0,0 +1,309 @@
...
+  try {
+    flags = toFlagsNumber(flags)
+  } catch (err) {
+    return cb(err)
```
mafintosh: nextTick

lib/fd.js
```diff
@@ -0,0 +1,309 @@
...
+
+    case 'a+' : return O_APPEND | O_CREAT | O_RDWR
+    case 'ax+': // Fall through.
+    case 'xa+': return O_APPEND | O_CREAT | O_RDWR | O_EXCL
+    case 'as+': // Fall through.
+    case 'sa+': return O_APPEND | O_CREAT | O_RDWR | O_SYNC
+  }
+
+  throw new errors.InvalidArgument(`Invalid value in flags: ${flags}`)
+}
```
mafintosh: the modularity in me screams make this func a separate module! :D :D :D

lib/storage.js
```diff
@@ -0,0 +1,38 @@
...
+    metadata: function (name) {
+      return storage(path.join(folder, 'metadata', name))
+    },
+    content: function (name) {
+      return storage(path.join(folder, 'content', name))
+    }
+  }
+}
+
+module.exports = {
```
mafintosh: module.exports = defaultStorage

test/fuzzing.js
```diff
@@ -0,0 +1,527 @@
...
+        seed: 'iteration #' + i,
+        debugging: false
+      })
+      await fuzz.run(100)
+    }
+    t.pass('fuzzing suceeded')
+  } catch (err) {
+    t.error(err, 'no error')
+  }
+})
```
mafintosh: IM LOVING THE FUZZ

mafintosh:
@andrewosh looking great! left a review but mostly nits. I'm thinking it's work moving back to `content`, `metadata` instead of `contentFeed`, `metadataFeed`. Whilst I like the latter, we reduce a ton of breakage by not changing them yet.

lib/fd.js
andrewosh: Nah the implicit copy is done inside the FUSE module -- this concat is to batch together small writes (increases FUSE performance). Think this should be removed or exposed as an opt?

README.md
andrewosh: Yep, added them, and `write`, back.

lib/fd.js
andrewosh: :+1:

lib/fd.js
andrewosh: Yep, updated to export only this function

lib/fd.js
andrewosh: :+1:

lib/storage.js
andrewosh: :+1:

test/fuzzing.js
andrewosh: #fuzzylife

andrewosh:
> @andrewosh looking great! left a review but mostly nits. I'm thinking it's work moving back to `content`, `metadata` instead of `contentFeed`, `metadataFeed`. Whilst I like the latter, we reduce a ton of breakage by not changing them yet.

Sounds good, changed the names back in the latest commits.

lib/fd.js
```diff
@@ -0,0 +1,309 @@
...
+  O_RDONLY,
+  O_WRONLY,
+  O_RDWR,
+  O_CREAT,
+  O_TRUNC,
+  O_APPEND,
+  O_SYNC,
+  O_EXCL
+} = fs.constants
+const O_ACCMODE = 3
```
mafintosh: is this not defined in fs.constants?

lib/storage.js
```diff
@@ -0,0 +1,34 @@
...
+      return storage.metadata(name, opts, self)
+    },
+    content: function (name, opts) {
+      return storage.content(name, opts, self)
+    }
+  }
+}
+
+module.exports = function defaultStorage (self, storage, opts) {
+  var folder = ''
```
mafintosh: let?

lib/fd.js
andrewosh: Done!

lib/fd.js
andrewosh: Got rid of the `fs` dependency (replaced it with `filesystem-constants`, as per offline discussion).

lib/storage.js
andrewosh: :+1:

mafintosh:
Rebased into master 💥 🎉

PR #234, opened 2019-02-07
Allow to add metadata to Stat entries

Frando:
This allows to add metadata to Stat entries. See the test for an example. 
(This PR is against the v10 branch)

### DIFF
--- a/README.md
+++ b/README.md
@@ -182,6 +182,7 @@ If `cached` is set to `true`, this function returns results only if they have al
 
 Write a file as a stream. Similar to fs.createWriteStream.
 If `options.cached` is set to `true`, this function returns results only if they have already been downloaded.
+`options.metadata` is optionally an object with string keys and buffer objects to set metadata on the file entry.
 
 #### `archive.writeFile(name, buffer, [options], [callback])`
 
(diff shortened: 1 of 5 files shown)


mafintosh:
@andrewosh @pfrazee this looked good to everyone right?

pfrazee:
@mafintosh Yeah the mechanics look good. This is going to be a very cool feature.

I'm a little concerned about what happens if lots of data is written into the metadata, which might cause stat()s to incur larger performance costs than expected. (Think: people storing icons.) What do you think about adding a byte limit?

mafintosh:
We def should add an advisory limit. We
can add that in another PR though unless someone wants to add that quickly before I merge :)

Frando:
Now that the hidden value support was added to hypertrie in hypertrie:#14 this PR should likely be reworked to use that api. The surface can either stay as it is so far (settings opts.metadata) or expose a new function instead (getMetadata, setMetadata). I propose to still keep the Map<name, bytes> schema for the metadata record so that different tools can use metadata without getting into conflicts.

Also, if I see things correctly, currently in hypertrie the hidden and non-hidden entries would have the same valueEncoding, right? If so then I'd propose to add support to hypertrie to set different valueEncoding for hidden and non-hidden nodes (so to have the regular hyperdrive Stat message for non-hidden nodes, and a Metadata message (with the name/bytes map) for the hidden nodes.

Opinions?

mafintosh:
@Frando Yes, we need to make valueEncoding configurable for each command. I do like the simplicity of storing this in the stat object though, but if you were to store *a ton* of metadata the hidden trie would be def be better

Frando:
I also thought about this some more and think that adding this is a good idea even with the hidden entries. Are there any TODOs left? For me it looks good.

Frando:
Can we get this in before the v10 release?

PR #235, opened 2019-03-15
Check if waiting for clear before looping

joehand:
Addressed the behavior seen in https://gist.github.com/joehand/ad9d3a3dc7f7d7ab611307d396025fbe. 

Not sure if this is fixing the CPU getting pegged issue...

Also probably related to the TODO in here =) and maybe a cleaner way to do this.


joehand:
See #236

PR #236, opened 2019-03-16
Fix for multiple callbacks for _fetchVersion

allain:


allain:
Fixes the same issue as pull request 10 but in a different way.

mafintosh:
Fixed in https://github.com/mafintosh/hyperdrive/commit/a1474b7986eec3d91620ed93bbc073e5eed704ff

PR #237, opened 2019-04-17
Update FD size during write

andrewosh:
This PR fixes a strange FUSE bug on OSX (`getattr` and `write` get called concurrently, to devastating effect...).

### DIFF
--- a/index.js
+++ b/index.js
@@ -588,7 +588,6 @@ class Hyperdrive extends EventEmitter {
           return cb(err)
         }
         if (this._writingFd && name === this._writingFd.path) {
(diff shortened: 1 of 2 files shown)


index.js
```diff
@@ -586,6 +587,10 @@ class Hyperdrive extends EventEmitter {
         } catch (err) {
           return cb(err)
         }
+        if (this._writingFd && name === this._writingFd.path) {
+          console.log('SETTING SIZE TO:', this._writingFd.stat.size)
```
mafintosh: accidental log

PR #240, opened 2019-06-27
Added extension support

RangerMauve:
Following up on [the hypercore extension support](https://github.com/mafintosh/hypercore/pull/214), this surfaces the extensions API in hyperdrive so that applications can build on top of it.

This could simplify code for things like Beaker's [datPeers](https://beakerbrowser.com/docs/apis/experimental-datpeers) API.

Adds:

- `extensions` option in constructor that gets passed to the metadata hypercore
- `extension` method for sending an extension message to all peers

TODO:

- [x] Tests
- [x] Docs

Should I maybe redo this PR against the v10 branch, or could we get it into a v9 release before then?

### DIFF
--- a/README.md
+++ b/README.md
@@ -78,6 +78,7 @@ Options include:
 {
   sparse: true, // only download data on content feed when it is specifically requested
   sparseMetadata: true // only download data on metadata feed when requested
+  extensions: [] // Optionally specify which extensions to use when replicating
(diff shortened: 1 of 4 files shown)


RangerMauve:
This is ready for review. Just need to wait for hypercore 7.5.0 to be released on NPM, then I'll need to update the package.json version

PR #242, opened 2019-07-23
Emit close event on archive close

RangerMauve:
It can be a little frustrating to manage hyperdrives (for example in the SDK) when you don't know when an application will be closing them.

This PR propagates the `close` event from the metadata hypercore to the hyperdrive.


mafintosh:
9.16.0

PR #243, opened 2019-07-23
Added sync-metadata event

RangerMauve:
This is useful for people to know when it's safe to use `readdir` and the such.

Currently people use a bunch of weird tricks to get it to work. 

This will make the tricks a little bit less arcane and somewhat documented. :P


RangerMauve:
Closing this for now. Downloading the full metadata is an antipattern.

Instead one should use `archive.metadata.update({ifAvailable: true}, whatever)`

PR #244, opened 2019-07-24
Fix readdir / on windows

RangerMauve:
Invoking `readdir('/')` before the metadata has loaded on Windows caused an error.

This is due to `path.resolve('/', '/')` returning `C:\\` and `unixify` converting that to `""` which makes `name === '/'` return false.

Using the posix `path` fixes this behavior.

Windows is weird. :P


t-mullen:
`path-browserify` doesn't have `path.posix`, so this change broke `hyperdrive` in the browser.

PR #248, opened 2019-08-08
Added events and docs

RangerMauve:
Adds the following:

- Added `extensions` array to constructor to be passed down to the metadata core
- `"extension"` event from the metadata feed to listen on extension messages
- `extension()` method for broadcasting extension messages to peers
- `.peers` getter for getting the current list of peers (useful for stuff like getting the health of an archive)
- `"peer-add"` event from the metadata feed so that actions can be taken when a peer is added (useful for extension stuff)
- `"peer-remove"` event from the metadata feed so that actions can be taken when a peer is removed
- `"close"` event so that actions can be taken when the archive is closed (useful for the SDK)
- docs for the `"update"` event so that people can be aware of it

Ideally these properties and events should be exposed in the RPC API in the daemon down the line, which is why I'd like to surface them at the top level rather than getting people to mess with the metadata feed directly.

TODO:

- Tests 😁

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,7 +1,7 @@
 # Hyperdrive
 
 #### *Note*: This is a prerelease version of Hyperdrive that's backed by [Hypertrie](https://github.com/mafintosh/hypertrie)
-#### This version is not yet API-complete. 
+#### This version is not yet API-complete.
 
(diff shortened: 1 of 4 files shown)


RangerMauve:
Woot! 💜💜

PR #250, opened 2019-08-28
add `archive.closeCheckout` method

poga:
fix https://github.com/mafintosh/hyperdrive/issues/241

### DIFF
--- a/index.js
+++ b/index.js
@@ -42,6 +42,7 @@ class Hyperdrive extends EventEmitter {
     this.live = true
     this.sparse = opts.sparse !== false
     this.sparseMetadata = opts.sparseMetadata !== false
+    this.listenersOnMetadata = {}
 
     this._corestore = defaultCorestore(storage, {
       ...opts,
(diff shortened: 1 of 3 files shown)


index.js
```diff
@@ -108,11 +109,17 @@ class Hyperdrive extends EventEmitter {
...
-    self.metadata.on('extension', extension)
-    self.metadata.on('peer-add', peeradd)
-    self.metadata.on('peer-remove', peerremove)
+    let listeners = {
+      'error': onerror,
+      'append': update,
+      'extension': extension,
+      'peer-add': peeradd,
+      'peer-remove': peerremove
+    }
```
mafintosh: @andrewosh I'm wondering fs there any need in a checkout to actually listen for these events?

poga: well if they're not needed for a checkout, that's even better! no need for the weird `closeCheckout()` method. :D

andrewosh: Yeah I'd agree that checkouts shouldn't be adding listeners for these events. We should prob do a check to see if the metadata matches what's passed in as an option (which is what a checkout does) before attaching listeners.

andrewosh:
Closing because this has been fixed in later releases.

PR #254, opened 2019-11-26
Added isInitiator in replication method

markopolojarvi:
isInitiator is required now so probably a good idea to add it to the documentation.


andrewosh:
Sorry for the super late review @markopolojarvi, but yes this needs to be in the README. Better late than never! I added your description manually in https://github.com/hypercore-protocol/hyperdrive/commit/4a10b3810c655a92b919cf4e9e927cc08322673c, rather than fixing up this PR to work with latest master.

PR #256, opened 2019-12-08
Fix path.posix not being available in browsers

t-mullen:
`path-browserify` does not provide a `.posix` implementation.

Environments without a posix implementation should just use the native implementation.


andrewosh:
Looks good to me. Thanks @t-mullen

PR #257, opened 2019-12-13
read replicated mount test

markopolojarvi:
This test tries to replicate hyperdrive with existing mount with files into another hyperdrive and read it. At the moment the replication fails.


andrewosh:
Closing this as per the comment in #258 

Big thanks again for including a tape test with the issue. Makes debugging much easier on my end

PR #261, opened 2020-01-13
fd parallel read fix

mafintosh:
fixes an issue with certain parallel reads from an fd

### DIFF
--- a/test/fd.js
+++ b/test/fd.js
@@ -354,3 +354,40 @@ tape('fd random-access write fails', function (t) {
     })
   })
 })
+
+tape('fd parallel reads', function (t) {
+  t.plan(3 * 2 + 1)
+
+  const drive = create()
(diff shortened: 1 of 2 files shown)


test/fd.js
```diff
@@ -354,3 +354,40 @@ tape('fd random-access write fails', function (t) {
...
 })
+
+tape('fd parallel reads', function (t) {
+  t.plan(3 * 2 + 1)
+
+  const drive = create()
+  const ws = drive.createWriteStream('test')
+
+  ws.write('foo')
+  setImmediate(function () {
```
andrewosh: Just curious why these all need to be setImmediate'd

mafintosh: To ensure that the stream is flushed (since we run in memory here setImmediate ensures all nextticks have run to completetion)

PR #262, opened 2020-01-13
FD close cancels pending reads

andrewosh:
When a file descriptor is closed, it should cancel all pending download ranges, which should in turn trigger cancellation errors for all pending reads. This is particularly important for FUSE, as it will prevent hanging terminals etc after a program exits before reads could complete.

### DIFF
--- a/test/fd.js
+++ b/test/fd.js
@@ -420,10 +420,8 @@ tape('fd close cancels pending reads', function (t) {
         })
         setImmediate(() => {
           // This should cancel the above read.
-          console.log('closing')
           clone.close(descriptor, err => {
             t.error(err, 'no error')
-            console.log('finished close')
             t.false(totalRead)
           })
         })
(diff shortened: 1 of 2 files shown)


test/fd.js
```diff
@@ -391,3 +391,43 @@ tape('fd parallel reads', function (t) {
...
+        // This read should hang without a timeout.
+        var totalRead = null
+        clone.read(descriptor, Buffer.allocUnsafe(64), 1024, 0, (err, total, buf) => {
+          t.true(err, 'read errored')
+          totalRead = total
+          t.end()
+        })
+        setImmediate(() => {
+          // This should cancel the above read.
+          console.log('closing')
```
mafintosh: stray log

test/fd.js
```diff
@@ -391,3 +391,43 @@ tape('fd parallel reads', function (t) {
...
+          clone.close(descriptor, err => {
+            t.error(err, 'no error')
+            console.log('finished close')
```
mafintosh: same

test/fd.js
andrewosh: oops

PR #264, opened 2020-01-22
add failing update-of-mountpoint test

mafintosh:


andrewosh:
Fixed in mountable-hypertrie@2.4.5. Merging in the test case.

PR #268, opened 2020-02-25
added destroy method

RangerMauve:
This adds a destroy method to hyperdrive which deletes its underlying data.

Same as the method in hypercore and [corestore](https://github.com/andrewosh/corestore/pull/9).

### DIFF
--- a/README.md
+++ b/README.md
@@ -389,3 +389,8 @@ Close a file. Similar to fs.close.
 
 Closes all open resources used by the archive.
 The archive should no longer be used after calling this.
+
+#### `archive.destroy([callback])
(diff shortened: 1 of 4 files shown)


pfrazee:
Does this destroy just the archive's data or all of corestore's data?

RangerMauve:
@pfrazee Depends on whether the archive is using a namespace.

If it's using a namespace, then only the cores that it's used will be destroyed. If it's not using a namespace and is the "root" archive, then everything will be destroyed.

RangerMauve:
Closing in favour of #271

PR #269, opened 2020-03-12
rework getContent to fix race condition and simplify

mafintosh:
Changes _getContent to map from feed instead of db. Makes checkout much easier. Also fixes a race condition if _getContent is ran many times before it finishes. Additionally it tries to load content in ready if it's cached locally, wheather or not it is writable.

Also fixes some bugs in test/mount that wasn't triggered before with namespaces being reused.


mafintosh:
@RangerMauve that was the upcoming change i mentioned in irc

PR #270, opened 2020-03-12
aggresively load content feed

mafintosh:
improves the latency bench by ~200-250 ms


andrewosh:
LGTM. Published 10.8.20

PR #277, opened 2020-06-02
Adding a mirror method

andrewosh:
Right now `download` is used to download the latest versions of specific files/directories within a drive. It'd be possible to extend this to support downloading all previous versions of those files/directories as well, but doing so might complicate the usage (i.e. `drive.download('/my-directory', { allVersions: true})`.

The UX there gets even trickier when considering mounts. What if `/my-directory` was previously a mount? It can get hard to reason about.

This PR adds a `mirror` method that does a full sync of a drive's metadata/content at the Hypercore level. It currently syncs all mounts as well, but this can be made configurable in the future. 

Any thoughts on introducing an additional method for this behavior?

### DIFF
--- a/index.js
+++ b/index.js
@@ -935,7 +935,7 @@ class Hyperdrive extends Nanoresource {
     // Getting all the mounts will trigger the event listeners above.
     this.getAllMounts({ content: true }, (err, mounts) => {
       if (err) return this.emit('error', err)
-      for (const [, { metadata, content }] of mounts) {
+      for (const { metadata, content } of mounts.values()) {
         oncore(metadata)
         oncore(content)
       }
@@ -957,7 +957,7 @@ class Hyperdrive extends Nanoresource {
     function oncore (core) {
       if (!core) return
       if (!self._mirrorRanges || self._mirrorRanges.has(core)) return
-      self._mirrorRanges.set(core, core.download())
+      self._mirrorRanges.set(core, core.download({ start: 0, end: -1 }))
     }
   }
 
--- a/index.js
+++ b/index.js
@@ -72,7 +72,7 @@ class Hyperdrive extends Nanoresource {
     this._fds = []
     this._writingFds = new Map()
     this._unlistens = []
-    this._mirrorRanges = null
+    this._unmirror = null
 
     this._metadataOpts = {
       key,
@@ -927,8 +927,8 @@ class Hyperdrive extends Nanoresource {
 
   mirror () {
     const self = this
-    if (this._mirrorRanges) return unmirror
-    this._mirrorRanges = new Map()
+    if (this._unmirror) return this._unmirror
+    const mirrorRanges = new Map()
 
     this.on('content-feed', oncore)
     this.on('metadata-feed', oncore)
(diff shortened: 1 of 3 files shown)


index.js
```diff
@@ -901,6 +902,42 @@ class Hyperdrive extends Nanoresource {
...
+    const self = this
+    if (this._mirrorRanges) return unmirror
+    this._mirrorRanges = new Map()
+
+    this.on('content-feed', oncore)
+    this.on('metadata-feed', oncore)
+    // Getting all the mounts will trigger the event listeners above.
+    this.getAllMounts({ content: true }, (err, mounts) => {
+      if (err) return this.emit('error', err)
+      for (const [, { metadata, content }] of mounts) {
```
mafintosh, line 915: .values() to avoid the array

index.js
```diff
@@ -901,6 +902,42 @@ class Hyperdrive extends Nanoresource {
...
 
+  mirror () {
+    const self = this
+    if (this._mirrorRanges) return unmirror
+    this._mirrorRanges = new Map()
+
+    this.on('content-feed', oncore)
+    this.on('metadata-feed', oncore)
+    // Getting all the mounts will trigger the event listeners above.
+    this.getAllMounts({ content: true }, (err, mounts) => {
```
mafintosh, line 913: Does getAllMounts reutn the metadata/content of the target drive here as well? Otherwise we need to add them if the events above have fired earlier

index.js
```diff
@@ -901,6 +902,42 @@ class Hyperdrive extends Nanoresource {
...
+      self.removeListener('metadata-feed', oncore)
+      for (const [ core, range ] of ranges) {
+        core.undownload(range)
+      }
+    }
+
+    function oncore (core) {
+      if (!core) return
+      if (!self._mirrorRanges || self._mirrorRanges.has(core)) return
+      self._mirrorRanges.set(core, core.download())
```
mafintosh, line 937: `core.download({ start: 0, end: -1 })`, otherwise it only gets until current feed.length

index.js
andrewosh: Yeah you mean the top-level drive? Yeah those will be included

index.js
andrewosh: Fixed

index.js
andrewosh: Fixed

index.js
mafintosh: 👍

index.js
```diff
@@ -901,6 +902,42 @@ class Hyperdrive extends Nanoresource {
     }
   }
 
+  mirror () {
+    const self = this
+    if (this._mirrorRanges) return unmirror
```
mafintosh, line 907: if you call mirror twice and call the 2nd unmirror, then the event listeneners attached from the first are never cleaned up

mafintosh: instead of storing mirrorRanges as state, mb store unmirror itself to simplify that?

index.js
```diff
@@ -918,23 +918,23 @@ class Hyperdrive extends Nanoresource {
...
+      for (const [ core, range ] of mirrorRanges) {
         core.undownload(range)
       }
     }
 
     function oncore (core) {
       if (!core) return
-      if (!self._mirrorRanges || self._mirrorRanges.has(core)) return
-      self._mirrorRanges.set(core, core.download({ start: 0, end: -1 }))
+      if (!self._unmirror || mirrorRanges.has(core)) return
```
mafintosh, line 936: You need || self._unmirror !== unmirror here as well to avoid a race condition

index.js
andrewosh: Fixed

index.js
andrewosh: Yep added this

todrobbins:
[“You did it...”](https://www.youtube.com/watch?v=gXBJRz_33Y8)

PR #278, opened 2020-06-03
Remove listeners on close

andrewosh:
The metadata/content feeds might not be closed when the drive is closed (they might be mounted elsewhere, or being passively replicated in the corestore). If we don't remove these listeners, the drive state will be leaked after close.

### DIFF
--- a/index.js
+++ b/index.js
@@ -133,6 +133,14 @@ class Hyperdrive extends Nanoresource {
       self.metadata.on('peer-add', peeradd)
       self.metadata.on('peer-remove', peerremove)
 
+      this.once('close', () => {
+        self.metadata.removeListener('error', onerror)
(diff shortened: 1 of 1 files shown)


index.js
```diff
@@ -121,18 +122,25 @@ class Hyperdrive extends Nanoresource {
         feed: this.metadata,
         sparse: this.sparseMetadata
       })
-      this.db.on('hypertrie', trie => {
-        this.emit('metadata-feed', trie.feed)
-        this.emit('mount', trie)
-      })
-      this.db.on('error', err => this.emit('error', err))
+      this.db.on('hypertrie', onhypertrie)
+      this.db.on('error', onerror)
```
mafintosh, line 126: We already have onerror?

mafintosh: Ah i see below

PR #280, opened 2020-06-11
Fix mountable-hypertrie min version

### LINKED ISSUE #279: Latest release is broken because this.db.close is not a function
When closing hyperdrives, I get immediate errors because `this.db.close` is not a function. Reason is that `mountable-hypertrie` is still on 2.4 but should likely be 2.6 where it was moved to inherit from nanoresource.

Frando:
Fixes #279


andrewosh:
Published 10.12.3

PR #285, opened 2020-07-10
Add docs for drive.watch

RangerMauve:
Also clarified that the update event doesn't get emitted unless you're watching files or downloading data.

### DIFF
--- a/README.md
+++ b/README.md
@@ -149,7 +149,7 @@ Emitted when a critical error during load happened.
 
 #### `drive.on('update')`
 
-Emitted when there is a new update to the drive.
+Emitted when there is a new update to the drive. Not triggered unless you load the data. Use `drive.watch()` to start loading.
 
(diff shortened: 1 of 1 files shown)


RangerMauve:
@andrewosh How does this README change look? :o I think it'd be handy for people trying to listen on updates.

KyleAMathews:
Is there a way to get _what_ file changed? That's necessary when watching directories.

RangerMauve:
@KyleAMathews To get the file that changed you'll want to diff against the previous version since that isn't being tracked AFAIK. @mafintosh or @andrewosh might have more insight though.

I usually use this for diffs: https://github.com/pfrazee/diff-file-tree

andrewosh (approved):

andrewosh:
Thanks @RangerMauve, and sorry for missing this one!

PR #287, opened 2020-07-13
Changes for Hyperspace

andrewosh:
This PR contains a handful of changes to make it easier to work with Hyperspace.

Most importantly, it introduces a Promises API that can be accessed through `drive.promises` (similar to the `fs` module).

It also asyncifies a handful of Hypercore methods that were previously assumed to be sync. These were updated to be optionally-async in Hypercore, so as to be compatible with Hyperspace's `RemoteHypercore`s.


index.js
```diff
@@ -186,15 +188,17 @@ class Hyperdrive extends Nanoresource {
      * (Otherwise, we need to read the feed's metadata block first)
      */
     function restore (keyPair) {
-      self.db.ready(err => {
+      self.db.ready(async err => {
```
mafintosh, line 191: why async?

andrewosh: leftover from debugging

index.js
```diff
@@ -958,8 +966,7 @@ class Hyperdrive extends Nanoresource {
     function oncore (core) {
       if (!core) return
       if (!self._unmirror || self._unmirror !== unmirror || mirrorRanges.has(core)) return
-      mirrorRanges.set(core, core.download({ start: 0, end: -1 }))
-    }
+      mirrorRanges.set(core, core.download({ start: 0, end: -1 }))}
```
mafintosh, line 969: typo?

andrewosh: yup

PR #290, opened 2020-07-24
Add info method

andrewosh:
This adds an `info` method which will return mount information about a file (and potentially other information that doesn't fit nicely into a Stat object in the future).


andrewosh:
10.16.0

PR #291, opened 2020-08-04
feat: .clear a file from local cache

okdistribute:
Fixes https://github.com/hypercore-protocol/hyperdrive/issues/228#issuecomment-668678754

This calls `clear` under the hood for the given file, using the same walking tree as the hyperdrive.download function. 

If this is desired, I will also write a companion section for the README.


okdistribute:
closing in favor of https://github.com/hypercore-protocol/hyperdrive/pull/292

PR #292, opened 2020-08-04
feat: .clear a file from local cache

okdistribute:
Fixes https://github.com/hypercore-protocol/hyperdrive/issues/228#issuecomment-668678754

This calls `clear` under the hood for the given file, using the same walking tree as the hyperdrive.download function. 

If this is a desired appraoch, I will also write a companion section for the README and tests for directories/mounts

### DIFF
--- a/index.js
+++ b/index.js
@@ -993,12 +993,43 @@ class Hyperdrive extends Nanoresource {
     }
   }
 
+  clear (path, opts, cb) {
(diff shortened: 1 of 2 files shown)


andrewosh:
Extracting that `_walk` function looks like a nice approach! LGTM

andrewosh:
Published 10.17.0

PR #298, opened 2020-09-12
L879 d1 supposed to be d2 ?

DougAnderson444:
Is L879 supposed to be d2 (and not d1, since d1 is not defined in this function)?


andrewosh (approved):

andrewosh:
Yep looks right. Thanks @DougAnderson444 and sorry for the delay

PR #299, opened 2020-09-24
remove duplicate import

anotherjesse:
pumpify was imported twice with two different names. removing the unused import


andrewosh:
Thanks @anotherjesse

Issue #301, opened 2020-10-07
Pass though opts parameter to metadata hypertrie


fsteff:
It would be useful to have the ability to pass through the `opts`  parameter  of `createWriteStream`, `createReadstream` etc. to the metadata db/hypertrie.
This e.g. would make it possible to create hidden files by simple passing the `hidden: true` option. 

However, according to @mafintosh this would result in internal clashes, therefore some namespacing is required (obviously, e.g. for the valueEncoding option).

So here's my suggestion for the options namespace:
```
{
  db: { /* opts passed to the db */ }
}
```

I'll try to implement it right away and create a pull request once it works ;-)

Discussion [on Discord](https://discordapp.com/channels/709519409932140575/709522092424429742/763132743923531817)

fsteff:
Possible limitations - the question here is if there would even be a usecase for it:
- usages of `_update `=> used by `setMetadata`/`removeMetadata`, which don't have the opts parameter at all
- usages of `_upsert `(what does that even do?)
- `symlink `does not have the opts parameter

PR #302, opened 2020-10-07
 Pass though opts parameter to metadata hypertrie

fsteff:
implements #301 using the opts.db namespace

But for hidden files to actually work as expected, [mountable-hypertrie#9](https://github.com/andrewosh/mountable-hypertrie/issues/9) has to be solved  - either in hyperdrive or there.


fsteff:
The newest commit uses my fork of mountable-hypertrie. Obviously this has to be changed back, but for demonstration purposes it works.

mafintosh:
Old major, thanks tho!

PR #306, opened 2021-01-07
Change to 'new' constructor pattern 

pfrazee:
Remains backwards-compat with function-call instantiation.

See https://github.com/hypercore-protocol/hyperbee/pull/19 for history

### DIFF
--- a/index.js
+++ b/index.js
@@ -1274,6 +1274,7 @@ class Hyperdrive extends Nanoresource {
 
 function HyperdriveCompat (...args) {
   if (!(this instanceof HyperdriveCompat)) return new HyperdriveCompat(...args)
+  Nanoresource.call(this)
   Hyperdrive.prototype._initialize.call(this, ...args)
 }
 Object.setPrototypeOf(HyperdriveCompat.prototype, Hyperdrive.prototype)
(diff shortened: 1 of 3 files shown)


pfrazee:
Do we want to do this with hypercore too?

index.js
```diff
@@ -1269,6 +1272,12 @@ class Hyperdrive extends Nanoresource {
   }
 }
 
+function HyperdriveCompat (...args) {
+  if (!(this instanceof HyperdriveCompat)) return new HyperdriveCompat(...args)
+  Hyperdrive.prototype._initialize.call(this, ...args)
```
mafintosh, line 1277: you need to call Nanoresource.call(this) before this line due to the inheritance

pfrazee: Good catch, sorry I missed that. Fixed

mafintosh: JS galore

mafintosh:
Looks good to me given green CI, will let @andrewosh get a chance to look at it also

andrewosh:
Now that's some JS. Looks good to me

andrewosh:
Published 10.20.0

RangerMauve:
Is the backwards compat going to be removed in the next major release?

mafintosh:
@RangerMauve i don't think so, but at some point, just because classes are a bit easier :)

PR #308, opened 2021-01-14
Add synchronous methods.

J-P-S-O:
This pull is in progress.
Just added synchronous `writeFileSync`

### DIFF
--- a/index.js
+++ b/index.js
@@ -575,12 +575,13 @@ class Hyperdrive extends Nanoresource {
       cb(null, opts.encoding && opts.encoding !== 'binary' ? buf.toString(opts.encoding) : buf)
     })
   }
+
   readFileSync(name, opts){
     if (typeof opts === 'string') opts = { encoding: opts }
     if (!opts) opts = {}
     name = fixName(name)
     collect(this.createReadStream(name, opts), function (err, bufs) {
-      if (err) err
+      if (err) throw err
       let buf = bufs.length === 1 ? bufs[0] : Buffer.concat(bufs)
       return opts.encoding && opts.encoding !== 'binary' ? buf.toString(opts.encoding) : buf
     })
--- a/index.js
+++ b/index.js
@@ -866,6 +866,7 @@ class Hyperdrive extends Nanoresource {
       self._del(name, cb)
     })
   }
+  
 
   replicate (isInitiator, opts) {
     // support replicate({ initiator: bool }) also
--- a/index.js
+++ b/index.js
@@ -826,6 +826,17 @@ class Hyperdrive extends Nanoresource {
       return cb(null, entries)
     })
   }
+  
+  readdirSync(name, opts){
+    if (!opts) opts = null
+    name = fixName(name)
+    const readdirStream = createReaddirStream(this, name, opts)
+    return collect(readdirStream, (err, entries) => {
+      if (err) throw err
+      return entries
+    })
+
+  }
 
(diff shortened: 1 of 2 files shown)


J-P-S-O:
Just added synchronous reading method.

J-P-S-O:
Added `readdirSync`

J-P-S-O:
Usage:
```javascript

let files = readdirSync("foo") // throws error if directory does not exist
let mydata = readFileSync("foo.bar")
writeFileSync("baz.txt", myBuffer) // should return null
```

J-P-S-O:
Ready for review.

mafintosh:
@J-P-S-O hi! did you test it? In general adding sync apis are hard as most things in Node are sync, such as streams, network etc (which is why we haven't added any)

J-P-S-O:
Well, the methods aren't very inovating, they just create the stream with all the callbacks and error handling hidden to the user.

J-P-S-O:
They depend on the async methods.

mafintosh:
@J-P-S-O yes, but you cannot return sync from a stream basically. If you try calling your readSync method here you'll see it just returns undefined as the sync function returns immediately.

index.js
```diff
@@ -576,6 +576,18 @@ class Hyperdrive extends Nanoresource {
...
+    if (typeof opts === 'string') opts = { encoding: opts }
+    if (!opts) opts = {}
+    name = fixName(name)
+    collect(this.createReadStream(name, opts), function (err, bufs) {
+      if (err) throw err
+      let buf = bufs.length === 1 ? bufs[0] : Buffer.concat(bufs)
+      return opts.encoding && opts.encoding !== 'binary' ? buf.toString(opts.encoding) : buf
+    })
+
+  }
```
J-P-S-O, line 589: Here, for example, the function main logic is similar to the async method, but error handling and treatment of the results are treated differently.

J-P-S-O:
I'm performing the tests.

index.js
mafintosh: Your return above just returns inside the callback function of collect so if you tried running this method it would always immediately return undefined

J-P-S-O:
I'll try to solve it.

index.js
J-P-S-O: Yes, I just saw it.

andrewosh:
Hey @J-P-S-O,

Appreciate the work you're putting into this, but the fundamental issue is that all underlying operations inside of Hyperdrive are async. The way you're approaching the sync methods really isn't possible. On the read side, the data isn't available sync, so you can never return sync. On the write side, if you return sync you're going to be suppressing errors or throwing them in the wrong place. 

In this way, Hyperdrive isn't analagous to Node's `fs` module, because we can't do blocking IO.

I think to accomplish what you're looking for, you can use the Promises interface on `drive.promises`, so you can `await drive.promises.readFile(...)`.

J-P-S-O:
Hi @andrewosh, thanks for the tip!

andrewosh:
Ah I'm sorry @J-P-S-O think there was a misunderstanding: I don't think we should have sync methods on Hyperdrive at all.

What I meant in my previous comment was that the usage you're looking for (calling Hyperdrive methods without dealing with callbacks) might already be covered by the Promises interface (`drive.promises`).

J-P-S-O:
Ok @andrewosh , now I understood it. I'm closing this Pull request.

Issue #309, opened 2021-02-25
Calling download() with no path: TypeError: expected a string


4c656554:
**Expected:** calling `download()` with no path should download all files. Presently getting following error (node v14.15.5, ubuntu):

**Actual:** 

```
const Hyperdrive = require('hyperdrive')

let d = new Hyperdrive('hyper://778f8d955175c92e4ced5e4f5563f69bfec0c86cc6f670352c457943666fe639')

d.download()
```

Gives the following error:

```
/myPath/node_modules/normalize-path/index.js:13
    throw new TypeError('expected a string');
    ^

TypeError: expected a string
    at normalizePath (/myPath/node_modules/normalize-path/index.js:13:11)
    at unixify (/myPath/node_modules/unixify/index.js:13:14)
    at fixName (/myPath/node_modules/hyperdrive/index.js:1292:10)
    at HyperdriveCompat.lstat (/myPath/node_modules/hyperdrive/index.js:712:12)
    at HyperdriveCompat.stat (/myPath/node_modules/hyperdrive/index.js:741:10)
    at HyperdriveCompat.stat (/myPath/node_modules/hyperdrive/index.js:738:49)
    at HyperdriveCompat._walk (/myPath/node_modules/hyperdrive/index.js:1057:10)
    at HyperdriveCompat.download (/myPath/node_modules/hyperdrive/index.js:1040:17)
    at Object.<anonymous> (/myPath/reptest.js:6:3)
```

From the docs:

> If no path is specified this will download all files.

https://github.com/hypercore-protocol/hyperdrive#drivedownloadpath-callback

4c656554:
I think a fix could be to check for undefined in line 1039 of index.js:

`if (path === undefined) path = '/'`

mixmix:
that or change the docs to say:

> to download all files set path to `/`

mixmix:
Sweet!

PR #310, opened 2021-02-25
Fix for "expected string" error in later Fixname call, Issue #309

4c656554:
Issue #309

### DIFF
--- a/index.js
+++ b/index.js
@@ -1036,7 +1036,9 @@ class Hyperdrive extends Nanoresource {
     opts.off = (feed, range) => {
       return feed.undownload(range)
     }
-
+    
+    if (path === undefined) path = '/'
+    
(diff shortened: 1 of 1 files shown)


andrewosh:
@4c656554 thanks for this, and sorry for being slow to review. Mind adding a test case for this, especially one to make sure that the callback is still handled correctly?

4c656554:
@andrewosh Thanks, correct me if I'm wrong, but I don't think it's possible (or it is possible then it'd be pretty strange) to call  `download` with path undefined *and* with a callback.

Perhaps a better fix would be to change the function def:

`  download (path="", opts, cb) { ...`

If you agree, I'll create a new commit with that?

Also, I've suggested a test case based on the directory download case, see #312

PR #311, opened 2021-03-03
recursive rmdir

heapwolf:
In this PR I added tests for `rmdir` (there were none). I also allow an options object to be threaded through `rmdir` to the `readdirIterator`. This allows us to do...

```js
drive.rmdir('beep', { recursive: true }, cb)

await drive.promises.rmdir('/boop', { recursive: true })
```


pfrazee:
I'm not conversant with this codebase but the PR includes tests and seems fine, @mafintosh or @andrewosh can you give a quick review or comment?

mafintosh:
Looks fine to me. Obvs not atomic but neither is it on a file system

Raynos (approved):
I think we have another bugfix for this where recursive did not work on empty directories.

andrewosh:
Hey @Raynos @heapwolf, you mentioned there might be a few more things to add to this PR before it's ready to merge? From the state of things currently, it all looks ready to go.

@heapwolf mentioned adding `cp` and `mv` to the API. `cp` definitely would be a good addition which we're happy to add, but `mv` has issues that have blocked us in the past. We'll want the moves to be atomic, and Hypertrie doesn't support atomically renaming directories.

Could we get your changes merged without `mv`?

PR #312, opened 2021-04-09
Adding a test case with no path as path argument

4c656554:
If no path is provided, the assumption is to download all.


4c656554:
Re #310

heapwolf (approved):

mafintosh:
Old major, thanks tho!

PR #317, opened 2021-06-15
allow content options as init argument (for use with custom encoding)

m4gpi:
enable `contentOpts` as an argument, and merge with the `contentOptions` function. This brings kappa-drive up to date with the latest hyperdrive, let me know if this is acceptable!

following discussion here from a while back: https://github.com/hypercore-protocol/hyperdrive/pull/289


mafintosh:
Old major, thanks tho!

PR #318, opened 2021-06-16
Update storage.js

mixmix:
defaultCorestore didn't have an "else" case, and it' also assume unix directory seperators "/".
this patch would fix those problems


heapwolf (approved):

mixmix:
Thanks @heapwolf

PR #319, opened 2021-07-30
Add metadata-download/upload and content-download/upload events

RangerMauve:
Listens for uploads and downloads in metadata and content and re-emits.

This is useful for apps that want to listen for changes in download progress without having to mess around with drive internals.


mafintosh:
10.21.0 :)

PR #325, opened 2021-12-10
Temp fix sparse windows + Passing settings to corestore(bug fix)

JocularMarrow:
Temporary fix to enable sparse on windows.

Also not sure about the opts but saw that they where not passed to the corestore when `case: 'string'`. So passed the options there also.
before
`return new Corestore(path => raf(join(storage, path), opts))`
now
`return new Corestore((path) => raf(join(storage, path), { ...opts, sparse: sparseWin }), opts);`

And if you are wondering about the spread operator I use it for overwriting the sparse only in the `raf()`. So it don't mess with other stuff when option `sparseWin` is true. Not sure if needed but wanted to be on the safe side.

### DIFF
--- a/lib/storage.js
+++ b/lib/storage.js
@@ -1,18 +1,32 @@
-const raf = require('random-access-file')
-const Corestore = require('corestore')
-const join = require('path').join
+const raf = require('random-access-file');
+const Corestore = require('corestore');
+const join = require('path').join;
+const fsctl = require('fsctl');
 
-module.exports = function defaultCorestore (storage, opts) {
-  if (isCorestore(storage)) return storage
-  
+module.exports = function defaultCorestore(storage, opts) {
+  if (isCorestore(storage)) return storage;
+  let factory = null;
+  console.log(storage);
   switch (typeof storage) {
-    case 'function': return new Corestore(path => storage(path), opts)
-    case 'string': return new Corestore(path => raf(join(storage, path), opts))
+    case 'function':
+      return new Corestore((path) => storage(path), opts);
+    case 'string':
+      const sparseWin =
+        opts.sparseWin && process.platform === 'win32'
+          ? fsctl.sparse
+          : opts.sparse;
+      return new Corestore(
+        (path) => raf(join(storage, path), { ...opts, sparse: sparseWin }),
+        opts
+      );
     default:
-      throw new Error('hyperdrive expects "storage" of type function|string, but got ' + typeof storage)
+      throw new Error(
+        'hyperdrive expects "storage" of type function|string, but got ' +
+          typeof storage
+      );
   }
-}
+};
 
-function isCorestore (storage) {
-  return storage.default && storage.get && storage.replicate && storage.close
+function isCorestore(storage) {
+  return storage.default && storage.get && storage.replicate && storage.close;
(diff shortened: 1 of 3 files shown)


lib/storage.js
```diff
@@ -1,18 +1,32 @@
...
+const join = require('path').join;
+const fsctl = require('fsctl');
 
-module.exports = function defaultCorestore (storage, opts) {
-  if (isCorestore(storage)) return storage
-  
+module.exports = function defaultCorestore(storage, opts) {
+  if (isCorestore(storage)) return storage;
+  let factory = null;
+  console.log(storage);
```
mafintosh, line 9: remove this log

lib/storage.js
```diff
@@ -1,18 +1,32 @@
...
+      return new Corestore((path) => storage(path), opts);
+    case 'string':
+      const sparseWin =
+        opts.sparseWin && process.platform === 'win32'
+          ? fsctl.sparse
+          : opts.sparse;
+      return new Corestore(
+        (path) => raf(join(storage, path), { ...opts, sparse: sparseWin }),
+        opts
+      );
```
mafintosh, line 21: remove all semicolons in general

mafintosh: And maybe add a comment saying this is a temp workaround for windows sparse files until it lands as the default here.

lib/storage.js
```diff
@@ -1,18 +1,32 @@
...
-      throw new Error('hyperdrive expects "storage" of type function|string, but got ' + typeof storage)
+      throw new Error(
+        'hyperdrive expects "storage" of type function|string, but got ' +
+          typeof storage
+      )
   }
 }
 
-function isCorestore (storage) {
+function isCorestore(storage) {
```
mafintosh, line 30: keep the space

lib/storage.js
```diff
@@ -1,18 +1,32 @@
 const raf = require('random-access-file')
 const Corestore = require('corestore')
 const join = require('path').join
+const fsctl = require('fsctl')
 
-module.exports = function defaultCorestore (storage, opts) {
+module.exports = function defaultCorestore(storage, opts) {
```
mafintosh, line 6: needs a space after the function name as well

lib/storage.js
```diff
@@ -1,18 +1,32 @@
 const raf = require('random-access-file')
 const Corestore = require('corestore')
 const join = require('path').join
+const fsctl = require('fsctl')
```
mafintosh, line 4: we need to make this browser friendly also, load it in a try catch and make sure to only use it if set below, ie

```
let fsctl = null
try {
  fsctl = require('fsctl')
} catch {
  // not supported, ie browser or unsupported platform - just skip
}

File: lib/storage.js [hunk 6]
```diff
@@ -1,18 +1,24 @@
 const raf = require('random-access-file')
 const Corestore = require('corestore')
 const join = require('path').join
-const fsctl = require('fsctl')
 
-module.exports = function defaultCorestore(storage, opts) {
+let fsctl = null;
```
ralphtheninja, line 5: One more semicolon.

ralphtheninja (changes requested):

PR #333, opened 2023-05-25
Use subEncoder for files sub

HDegroote:
Original here: https://github.com/holepunchto/hyperdrive-next/pull/57 

Marked as draft because I'm not sure exactly which parts of the code related to batches were work-arounds and which were on purpose. Note in particular that I am now passing the blobs through as a parameter instead of loading it again from this.db in a batch, which enables me to use the _db arg to pass the batch and to get rid of the _files arg (I couldn't figure out why a batch needed to reload the blobs, but there might have been a reason).

There's also a TODO marked in the PR to verify something in the close logic, where we recently made a fix with comment 'workaround to flush the batches for now.' => I think that's no longer relevant, but sanity checking that before fully removing

### DIFF
--- a/index.js
+++ b/index.js
@@ -99,8 +99,6 @@ module.exports = class Hyperdrive extends ReadyResource {
         await this.blobs.core.close()
       }
       await this.db.close()
-      // TODO:verify that the reason for this workaround is now taken away
-      // await this.files.close() // workaround to flush the batches for now. TODO: kill the sub!
     } catch (e) {
       safetyCatch(e)
     }
--- a/index.js
+++ b/index.js
@@ -277,14 +277,17 @@ module.exports = class Hyperdrive extends ReadyResource {
 
   diff (length, folder, opts = {}) {
     if (typeof folder === 'object' && folder && !opts) return this.diff(length, null, folder)
+
+    const range = {}
     if (folder) {
       if (folder.endsWith('/')) folder = folder.slice(0, -1)
       if (folder) folder = normalizePath(folder)
-      opts = { gt: folder + '/', lt: folder + '0', ...opts }
+      range.gt = folder + '/'
+      range.lt = folder + '0'
     }
 
     opts = { ...opts, keyEncoding: FILES_SUB }
-    return this.db.createDiffStream(length, opts)
+    return this.db.createDiffStream(length, range, opts)
   }
 
   async downloadDiff (length, folder, opts) {
--- a/index.js
+++ b/index.js
@@ -19,18 +19,19 @@ module.exports = class Hyperdrive extends ReadyResource {
       opts = key
       key = null
     }
-    const { _checkout, _db, onwait, _blobs } = opts
+    const { _checkout, _db, onwait, _batch } = opts
     this._onwait = onwait || null
 
     this.corestore = corestore
     this.db = _db || makeBee(key, corestore, this._onwait)
+    this._batch = _batch
 
-    this.blobs = _blobs || null
+    this.blobs = null
     this.supportsMetadata = true
 
     this._openingBlobs = null
     this._checkout = _checkout || null
-    this._batching = !!_blobs
+    this._batching = !!_batch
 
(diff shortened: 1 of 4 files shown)


index.js
```diff
@@ -8,6 +8,9 @@ const SubEncoder = require('sub-encoder')
 const ReadyResource = require('ready-resource')
 const safetyCatch = require('safety-catch')
 
+const DB_KEY_ENCODING = 'utf-8'
+const FILES_SUB = (new SubEncoder()).sub('files', DB_KEY_ENCODING)
```
mafintosh, line 12: Just do new SubEncoder(files, enc) here

index.js
```diff
@@ -262,14 +275,16 @@ module.exports = class Hyperdrive extends ReadyResource {
...
     if (typeof folder === 'object' && folder && !opts) return this.diff(length, null, folder)
     if (folder) {
       if (folder.endsWith('/')) folder = folder.slice(0, -1)
       if (folder) folder = normalizePath(folder)
       opts = { gt: folder + '/', lt: folder + '0', ...opts }
     }
-    return this.files.createDiffStream(length, opts)
+
+    opts = { ...opts, keyEncoding: FILES_SUB }
+    return this.db.createDiffStream(length, opts)
```
mafintosh, line 287: refactor this to use the range arg. then the gt/lt stuff is not mixed in

index.js
```diff
@@ -310,8 +325,9 @@ module.exports = class Hyperdrive extends ReadyResource {
     await Promise.allSettled(proms)
   }
 
-  entries (opts) {
-    return this.files.createReadStream(opts)
+  entries (opts = {}) {
```
mafintosh, line 328: lets make this (range, opts) like readstream in bee now

index.js
```diff
@@ -525,7 +545,7 @@ function makeBee (key, corestore, onwait) {
     : { name: 'db', cache: true, onwait }
   const core = corestore.get(metadataOpts)
   const metadata = { contentFeed: null }
-  return new Hyperbee(core, { keyEncoding: 'utf-8', valueEncoding: 'json', metadata })
+  return new Hyperbee(core, { keyEncoding: DB_KEY_ENCODING, valueEncoding: 'json', metadata })
```
mafintosh, line 548: would just inline the db_key_encoding personally

index.js
HDegroote: Went with

```
  entries (range = {}, opts = {}) {
    if (opts) opts.keyEncoding = FILES_SUB
    else range.keyEncoding = FILES_SUB // backward compat

    return this.db.createReadStream(range, opts)
  }

```
to keep backward compat

index.js
HDegroote: Went with the constant because we also need it when defining `FILES_SUB`, but fine with explicitly writing `utf-8` for both too

HDegroote:
Also pushed another commit to refactor `watch()`, so it also uses `FILES_SUB` instead of defining its own subEnc

HDegroote:
This PR has a bug when creating a batch on a non-ready core--looking into it.

HDegroote:
Fixed the bug in a not-so-elegant way, by always reloading the blobs (as before this PR)

The current flow for a batch is:
- Pass the original db, as well as the newly created batch to the new Hyperdrive instance
- in drive._open:
  - Load the blobs core, using the original db (needed to access the header)
  - Set this.db to the passed-in batch, so all operations on the db now operate on the batch
  - 
(it's this last step which is pretty ugly, but not sure if to try to clean up the entire flow in this PR, or to leave that for another one)

LuKks:
@HDegroote Because a batch instance doesn't have `getHeader` that's the issue, right? The easy fix is to hack it around and use `batch.tree` property when it's a batch, or to add the method to the Batch class

LuKks:
I moved the branch to the repo itself while preserving your commits so I could do some debugging:
https://github.com/holepunchto/hyperdrive/pull/344

mafintosh:
landed in main

PR #337, opened 2023-05-26
Fix blobs session leak on batching

HDegroote:
Currently a new blobs core session is created whenever a new batch is created, but they are never closed. This PR always closes the blobs core of a batch.

### DIFF
--- a/index.js
+++ b/index.js
@@ -86,27 +86,23 @@ module.exports = class Hyperdrive extends ReadyResource {
     })
   }
 
-  flush () {
-    return this.files.flush()
+  async flush () {
+    await this.files.flush()
+    return this.close()
   }
 
   async _close () {
-    if (this._batching) {
-      await this.blobs.core.close()
-      return this.files.close()
-    }
-
     try {
       if (this.blobs !== null && (this._checkout === null || this.blobs !== this._checkout.blobs)) {
         await this.blobs.core.close()
       }
-      await this.db.close()
+      if (!this._batching) await this.db.close()
       await this.files.close() // workaround to flush the batches for now. TODO: kill the sub!
     } catch (e) {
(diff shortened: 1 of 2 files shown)


index.js
```diff
@@ -86,23 +86,23 @@ module.exports = class Hyperdrive extends ReadyResource {
...
-  flush () {
-    return this.files.flush()
+  async flush () {
+    await this.files.flush()
+    return this.close()
   }
 
   async _close () {
-    if (this._batching) return this.files.close()
     try {
```
LuKks, line 95: We should replace the try-catch for a `.catch(safetyCatch)`, much better because if we're already ignoring errors then at least we should ignore them individually per async call

I mean like this:
```js
const closeBlobs = this.blobs !== null && (this._checkout === null || this.blobs !== this._checkout.blobs)

if (closeBlobs) await this.blobs.core.close().catch(safetyCatch)
if (!this._batching) await this.db.close().catch(safetyCatch)
await this.files.close().catch(safetyCatch)
```
And if somehow we can make the `closeBlobs` thing clearer lol

LuKks: Benefit is that it allows to other things to try close as well, maybe not that much of a benefit but it's something

LuKks: Or just a try-catch to the whole function

index.js
```diff
@@ -86,23 +86,23 @@ module.exports = class Hyperdrive extends ReadyResource {
     })
   }
 
-  flush () {
-    return this.files.flush()
+  async flush () {
+    await this.files.flush()
+    return this.close()
```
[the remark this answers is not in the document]
mafintosh: yea...

index.js
mafintosh: i think this is ok, and clear

PR #338, opened 2023-05-27
Add aliases for core properties

LuKks:


LuKks:
In combination with https://github.com/holepunchto/hyperbee/pull/119

It potentially allows a simpler API for replication like this:
```js
const swarm = new Hyperswarm()
swarm.on('connection', (socket) => drive.replicate(socket))
swarm.join(drive) // Or a bee
```
It would always pass down `{ discoveryKey, findingPeers }` for all building blocks like Hypercore, Hyperbee, and Hyperdrive. User should still manually call the replicate method to avoid any magic, but looks very safe to automatically handle the discovery key to find peers.

mafintosh:
11.1.0

PR #341, opened 2023-06-02
Add drive.compare(entryA, entryB)

LuKks:
Should be similar or very "interoperable" with https://github.com/holepunchto/localdrive/pull/25

### DIFF
--- a/index.js
+++ b/index.js
@@ -218,7 +218,8 @@ module.exports = class Hyperdrive extends ReadyResource {
   }
 
   compare (a, b) {
-    return a.seq - b.seq
+    const diff = a.seq - b.seq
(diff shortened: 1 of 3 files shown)


index.js
```diff
@@ -217,6 +217,10 @@ module.exports = class Hyperdrive extends ReadyResource {
     return this.files.del(normalizePath(name))
   }
 
+  compare (a, b) {
+    return a.seq - b.seq
```
LuKks, line 221: I think it should return `-1`, `0`, or `1`, then if Localdrive does the same then both API results are the same

LuKks: I just don't know if you actually need the difference in the output?

mafintosh: sure, up to you

PR #344, opened 2023-06-16
Remove files sub

LuKks:
Rebased from external Hans's branch to debug a bit

### DIFF
--- a/index.js
+++ b/index.js
@@ -256,7 +256,7 @@ module.exports = class Hyperdrive extends ReadyResource {
   }
 
   async entry (name, opts) {
-    if (typeof name !== 'string') return Promise.resolve(name)
+    if (typeof name !== 'string') return name
 
     return this.db.get(std(name), { ...opts, keyEncoding })
   }
--- a/index.js
+++ b/index.js
@@ -282,8 +282,8 @@ module.exports = class Hyperdrive extends ReadyResource {
 
     const range = {}
     if (folder) {
-      if (folder) folder = std(folder)
       if (folder.endsWith('/')) folder = folder.slice(0, -1)
+      if (folder) folder = std(folder)
       range.gt = folder + '/'
       range.lt = folder + '0'
     }
--- a/index.js
+++ b/index.js
@@ -498,7 +498,7 @@ function shallowReadStream (files, folder, keys) {
       let node = null
 
       try {
-        node = await files.peek({ gt: folder + prev, lt: folder + '0' }, { keyEncoding })
+        node = await files.peek(binaryRange(folder, prev), { keyEncoding })
       } catch (err) {
         return cb(err)
       }
@@ -542,8 +542,8 @@ function std (name, removeSlash) {
   return name
 }
 
-function binaryRange (name) {
+function binaryRange (name, prev = '/') {
   if (!name) return null
   // '0' is binary +1 of /
-  return { gt: name + '/', lt: name + '0' }
+  return { gt: name + prev, lt: name + '0' }
 }
--- a/index.js
+++ b/index.js
@@ -8,7 +8,7 @@ const SubEncoder = require('sub-encoder')
 const ReadyResource = require('ready-resource')
 const safetyCatch = require('safety-catch')
 
-const FILES_SUB = new SubEncoder('files', 'utf-8')
+const ENC = { keyEncoding: new SubEncoder('files', 'utf-8') }
 
 module.exports = class Hyperdrive extends ReadyResource {
   constructor (corestore, key, opts = {}) {
@@ -205,11 +205,11 @@ module.exports = class Hyperdrive extends ReadyResource {
   async put (name, buf, { executable = false, metadata = null } = {}) {
     await this.getBlobs()
     const blob = await this.blobs.put(buf)
-    return this.db.put(std(name), { executable, linkname: null, blob, metadata }, { keyEncoding: FILES_SUB })
+    return this.db.put(std(name), { executable, linkname: null, blob, metadata }, ENC)
   }
 
   async del (name) {
-    return this.db.del(std(name), { keyEncoding: FILES_SUB })
+    return this.db.del(std(name), ENC)
   }
 
   compare (a, b) {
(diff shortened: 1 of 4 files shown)


index.js
```diff
@@ -546,15 +536,18 @@ function shallowReadStream (files, folder, keys) {
...
-    : { name: 'db', cache: true, exclusive: true, onwait }
-  const core = corestore.get(metadataOpts)
-  const metadata = { contentFeed: null }
-  return new Hyperbee(core, { keyEncoding: 'utf-8', valueEncoding: 'json', metadata })
+function makeBee (key, corestore, opts) {
+  const name = key ? undefined : 'db'
+  const core = corestore.get({ key, name, cache: true, exclusive: true, onwait: opts.onwait })
+
+  return new Hyperbee(core, {
+    keyEncoding: 'utf-8',
```
LuKks, line 544: Maybe pass encoding here directly?

index.js
```diff
@@ -94,39 +93,31 @@ module.exports = class Hyperdrive extends ReadyResource {
...
+      await this.corestore.close().catch(safetyCatch)
     }
   }
 
   async _openBlobsFromHeader (opts) {
     if (this.blobs) return true
 
-    const header = await this.db.getHeader(opts)
+    const db = this.db.tree || this.db
+    const header = await db.getHeader(opts)
```
LuKks, line 120: Is that ok? If it's a batch then it will use the `tree` property to get the actual Hyperbee instance

LuKks:
Does this also needs like in watch eg `{ gt: encode(folder + '/'), lt: encode(folder + '0') }`:
https://github.com/holepunchto/hyperdrive/blob/7d3fbe52a5cd13365e7dcd8693bff1a3a527f849/index.js#L507

index.js
```diff
@@ -361,22 +346,20 @@ module.exports = class Hyperdrive extends ReadyResource {
...
 
-    if (folder.endsWith('/')) folder = folder.slice(0, -1)
-    if (folder) folder = normalizePath(folder)
+    folder = std(folder || '/', true)
+
+    if (opts && opts.recursive === false) return shallowReadStream(this.db, folder, false)
 
-    if (recursive === false) return shallowReadStream(this.files, folder, false)
     // '0' is binary +1 of /
     return folder ? this.entries({ gt: folder + '/', lt: folder + '0' }) : this.entries()
```
LuKks, line 357: Something to consider, we are saying "if folder then use range, otherwise default" but we are not doing that in watch for example

Can't just assume to always use a range?

LuKks: Since we are normalizing all filenames

Or watch should do the same? (explicit condition to use range)

index.js
```diff
@@ -172,7 +163,8 @@ module.exports = class Hyperdrive extends ReadyResource {
       await blobsCore.ready()
 
       this.blobs = new Hyperblobs(blobsCore)
-      this.db.metadata.contentFeed = this.blobs.core.key
+      const db = this.db.tree || this.db
+      db.metadata.contentFeed = this.blobs.core.key
```
LuKks, line 167: Same here about `.tree`, I think it's ok

index.js
mafintosh: Lets add getHeader on the batch instead

index.js
```diff
@@ -264,39 +254,34 @@ module.exports = class Hyperdrive extends ReadyResource {
...
+    folder = std(folder || '/', true)
 
-    const encoder = new SubEncoder()
-    const files = encoder.sub('files', this.db.keyEncoding)
-    const options = { map: (snap) => this._makeCheckout(snap) }
-
-    return this.db.watch({ gt: files.encode(folder + '/'), lt: files.encode(folder + '0') }, options)
+    return this.db.watch({
+      gt: ENC.keyEncoding.encode(folder + '/'),
+      lt: ENC.keyEncoding.encode(folder + '0')
```
mafintosh, line 275: why not just pass the keyEncoding?

index.js
mafintosh: i think the batch needs a getter for the metadata or a static hyperbee method that returns the db from either a batch or bee

LuKks: Setter*?

I think we're already in the path of making the Batch as similar as Hyperbee so a setter for the metadata makes sense. Same about adding `getHeader` method to Batch

So do I add that or we just leverage the `getBee` helper in Hyperdrive now?

index.js
LuKks: Addressing this in the other comment: https://github.com/holepunchto/hyperdrive/pull/344#discussion_r1233046936

index.js
```diff
@@ -94,39 +93,30 @@ module.exports = class Hyperdrive extends ReadyResource {
...
-    try {
-      if (this.blobs !== null && (this._checkout === null || this.blobs !== this._checkout.blobs)) {
-        await this.blobs.core.close()
-      }
-      if (!this._batching) await this.db.close()
-      await this.files.close() // workaround to flush the batches for now. TODO: kill the sub!
-    } catch (e) {
-      safetyCatch(e)
+    if (this.blobs && (!this._checkout || this.blobs !== this._checkout.blobs)) {
+      await this.blobs.core.close().catch(safetyCatch)
```
mafintosh, line 106: use try/catch

index.js
```diff
@@ -546,15 +520,30 @@ function shallowReadStream (files, folder, keys) {
...
+
+function std (name, removeSlash) {
+  name = unixPathResolve('/', name)
+  if (removeSlash && name.endsWith('/')) name = name.slice(0, -1)
+  return name
 }
 
-function normalizePath (name) {
-  return unixPathResolve('/', name)
+function binaryRange (name, prev = '/') {
```
mafintosh, line 545: prefixRange rather than binaryRange

index.js
```diff
@@ -94,39 +93,30 @@ module.exports = class Hyperdrive extends ReadyResource {
...
-      _files: this.files.batch()
+      _db: this.db.batch()
     })
   }
 
-  async flush () {
-    await this.files.flush()
-    return this.close()
+  flush () {
+    return this.db.flush()
```
HDegroote, line 101: Sanity checking: we no longer need to close this upon flushing a batch? Not really understanding why it's no longer needed, as `Hyperdrive.batch()` does return a new `Hyperdrive` object

Also sanity checking: it's normal that `batch()` returns a Hyperdrive object on the same `this.corestore` instance, without creating a new session? That seems wrong (definitely if we close the passed in `Hyperdrive` on flushing, as I think we should)

HDegroote: Just noticed that we conditionally close the corestore in `Hyperdrive.close()`, not closing it if we're in a checkout or batch, so calling close with the code as-is seems fine, it won't close the corestore in a batch.

Do note that this means that, when we close a Hyperdrive, any existing checkouts will become unusable (because the underlying corestore is closed), but that might be expected.

LuKks: I think you're right! The reason I didn't close it is because in main branch I didn't immediately see `this.files` being closed but it's actually in the close method so my bad!

Also, it should close it because MirrorDrive expects it as well:
https://github.com/holepunchto/mirror-drive/blob/cbd6dea2f175522536cb767f38fc6d23c9712677/index.js#L87
I mean, it does flush but it doesn't close, I think Mathias saw that so flush must close

Issue #342, opened 2023-06-03
Entry names with dashes break `drive.readdir`


josephmturner:
```
await drive.put('/foo', Buffer.from('foo'))
await drive.put('/foo-bar', Buffer.from('foo-bar'))

for await (const file of drive.readdir('/')) {
  console.log(file)
}
```
Expected:
```
foo
foo-bar
```
Actual:
```
foo
```

mafintosh:
fixed in latest

josephmturner:
Thank you!!

July 4, 2023 4:07 PM, "Mathias Buus" ***@***.*** ***@***.******@***.***>)> wrote:
	fixed in latest

PR #347, opened 2023-07-04
Test readdir filenames with dashes

LuKks:


LuKks:
Reproducible test for https://github.com/holepunchto/hyperdrive/issues/342

PR #348, opened 2023-07-07
Pass get options to Hyperbee and Hyperblobs

LuKks:


LuKks:
@mafintosh Ready!

index.js
```diff
@@ -272,7 +272,12 @@ module.exports = class Hyperdrive extends ReadyResource {
   async _entry (name, opts) {
     if (typeof name !== 'string') return name
 
-    return this.db.get(std(name, false), { ...opts, keyEncoding })
+    try {
+      return await this.db.get(std(name, false), { ...opts, keyEncoding })
+    } catch (err) {
+      if (err.message === 'Block not available locally') return null
```
mafintosh, line 278: update hyperbee to use the correct error from hyperdrive instead https://github.com/holepunchto/hypercore/blob/main/lib/errors.js#L47

mafintosh: actually this should NOT return null, thats ambigious. null means its not in the bee, thats why bee throws. but make bee use the error above instead so its typed

PR #351, opened 2023-07-26
Let get pass forward opts

HDegroote:
Relevant to be able to do things like `await drive.get('./file', { wait: false })`, so you only get the file if it's locally available


LuKks:
Related https://github.com/holepunchto/hyperdrive/pull/348

PR #355, opened 2023-08-05
namespace blob core when generated always

mafintosh:


mafintosh:
fully backwards compat as blob keys are preferred loading from the header

andrewosh (approved):

PR #359, opened 2023-10-26
Fix purge test

LuKks:


LuKks:
Unrelated, lots of space-unary-ops errors by Standard (probably a sub dep)

HDegroote:
Great find!

I do think the problem can be solved in a different way, by making all tests which need a tmp dir use a helper, and by ensuring that the helper cleans up fully before finishing. Then we don't have to work around unintended side effects in the next test
That would entail using the helper here:
https://github.com/holepunchto/hyperdrive/blob/97f9ab89d4bc14ea346d6c96ae968eca73ce701b/test.js#L79
https://github.com/holepunchto/hyperdrive/blob/97f9ab89d4bc14ea346d6c96ae968eca73ce701b/test.js#L132
https://github.com/holepunchto/hyperdrive/blob/97f9ab89d4bc14ea346d6c96ae968eca73ce701b/test.js#L1397

And updating the helper to explicitly await the teardown, like here:
```
function createTmpDir (t) {
  const tmpdir = path.join(os.tmpdir(), 'hyperdrive-test-')
  const dir = fs.mkdtempSync(tmpdir)
  t.teardown(async () => await fs.promises.rm(dir, { recursive: true }))
  return dir
}
```

We could even add a sanity check before the dir creation, to rm it if there was one from an earlier run (for example if it was aborted midway), genre `fs.rmSync('tmpdir', { recursive: true, force: true })`

What do you think?

HDegroote (approved):
Misunderstood the intent of this PR initially. It solves the edge-case of both the blobs and db core starting with the same 2 hex chars, thereby living in the same directory.

The PR looks good to me

PR #373, opened 2024-08-27
add get db key as a static method

rafapaezbas:
This PR adds a static method that returns the primary key of a Hyperdrive for a corestore instance.

### DIFF
--- a/index.js
+++ b/index.js
@@ -44,6 +44,14 @@ module.exports = class Hyperdrive extends ReadyResource {
     return this.entries()[Symbol.asyncIterator]()
   }
 
+  static async getDriveKey (corestore) {
+    const core = makeBee(undefined, corestore)
+    await core.ready()
+    const key = core.key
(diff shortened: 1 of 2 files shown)


index.js
```diff
@@ -44,6 +44,15 @@ module.exports = class Hyperdrive extends ReadyResource {
     return this.entries()[Symbol.asyncIterator]()
   }
 
+  static async getDbKey (corestore) {
```
mafintosh, line 47: getDriveKey seems a bit more correct

index.js
```diff
@@ -44,6 +44,15 @@ module.exports = class Hyperdrive extends ReadyResource {
...
+    const core = makeBee(undefined, corestore)
+    await core.ready()
+    const key = core.key
+    await core.close()
+    await corestore.close()
```
mafintosh, line 52: dont think it should close it since its a static method, but open for input

Issue #364, opened 2024-01-12
Add `rename` method


RangerMauve:
It'd be useful to have an easy way to move entries to new locations. At the moment one needs to either duplicate data and delete the old data in a batch, or mess with internal APIs.

I propose something like this:

```javascript
async rename(src, dst) {
  const entry = await this.entry(src)
  const batch = this.db.batch()
  batch.del(src)
  batch.put(dst, entry)
  batch.flush()
}
```

Would a PR that adds a working method as well as unit tests work for y'all?

PR #376, opened 2024-09-27
Add `drive.rename(oldPath, newPath)`

lejeunerenard:
Only updates the file entry in the hyperbee.

This PR was inspired by someone wanting to rename the file without having to re-append the blobstore entry.

### DIFF
--- a/index.js
+++ b/index.js
@@ -358,6 +358,12 @@ module.exports = class Hyperdrive extends ReadyResource {
     await Promise.all(proms)
   }
 
+  async rename (oldPath, newPath, opts) {
+    const existingEntry = await this.entry(oldPath, opts)
+    await this.db.del(std(oldPath, false), { keyEncoding })
+    return this.db.put(std(newPath, false), existingEntry.value, { keyEncoding })
+  }
+
   async symlink (name, dst, { metadata = null } = {}) {
     return this.db.put(std(name, false), { executable: false, linkname: dst, blob: null, metadata }, { keyEncoding })
   }
--- a/README.md
+++ b/README.md
@@ -31,18 +31,20 @@ console.log(entry) // => { seq, key, value: { executable, linkname, blob, metada
 
 await drive.del('/images/old-logo.png')
 
+await drive.rename('/images/blob.txt', '/images/example.txt')
+
 await drive.symlink('/images/logo.shortcut', '/images/logo.png')
 
 for await (const file of drive.list('/images')) {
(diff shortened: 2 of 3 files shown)


lejeunerenard:
Addresses #364. Should this renaming happen as a batch append? I had assumed that if batching was required, it could be done at the `drive` level instead, aka:
```js
const b = drive.batch()
await b.rename(oldPath, newPath)
await b.flush()
```

serapath:
@HDegroote any chance this could be merged? Maybe even soon? :-)

index.js
```diff
@@ -358,6 +358,12 @@ module.exports = class Hyperdrive extends ReadyResource {
     await Promise.all(proms)
   }
 
+  async rename (oldPath, newPath, opts) {
+    const existingEntry = await this.entry(oldPath, opts)
+    await this.db.del(std(oldPath, false), { keyEncoding })
+    return this.db.put(std(newPath, false), existingEntry.value, { keyEncoding })
```
HDegroote, line 364: This logic is prone to race conditions:
- If the process crashes/stops after doing the del, but before the put runs, the entry's metadata no longer exists
- if rename runs twice for the same oldPath and newPath, and the second one starts after the first one finished its del operation (but before it finished its put), then `existingEntry` will be null, so the second run will override the result of the first one and make it point to null

The first issue can be resolved by using a batch, and the second with some sort of lock or debounce, but I'm not sure it's worth the complexity for this feature

HDegroote:
> @HDegroote any chance this could be merged? Maybe even soon? :-)

It's @mafintosh  who decides on merging PR's for Hyperdrive, so best to ping him.

I reviewed the PR though. I think it's a less trivial feature than you initially thought, because of the potential for race conditions.

mafintosh:
Needs a batch, no need for extra lock. The read should be on the batch as well. Docs need to specify that its for a single entry also

index.js
lejeunerenard: Great catch! I should have thought of this.
Added `drive.batch()` to lock the drive while renaming which should address the race condition and the mid process crash concern.

lejeunerenard:
Specified that `rename()` applies to one entry and added a test to prove that renaming works on symlinks as well as blob entries.

serapath:
just wanted to come back and wonder if rename could be done using `.batch` to see further work has happened and it might work now?

The idea was to create a method `.rename` that just internally uses batch with all operations required to make rename work.

PR #380, opened 2024-12-13
Clean up dirs after test

HDegroote:
Currently we leak about 70 tmp dirs when running  `npm t` (mostly due to the one created in `testenv()` that wasn't being torn down.

With this PR, all temp dirs are cleaned up


mafintosh:
Can we just move the tmp dir generate to test-tmp instead?

HDegroote:
Accidentally pushed to the commit moving to test-tmp directly to main, so will close this PR

PR #385, opened 2025-03-06
feat: has method implementation

yuranich:
Use case: 
we have a key from hd running remotely and create new instance with local corestore

````
const drive = new Hyperdrive(store, key)
...
drive.download('/')
```

and tomorrow we do again

```
const drive = new Hyperdrive(store, key)
```

need a quick way to check if files already available locally ( and let user know).

### DIFF
--- a/index.js
+++ b/index.js
@@ -477,6 +477,20 @@ module.exports = class Hyperdrive extends ReadyResource {
     await Promise.allSettled(proms)
   }
 
+  hasBlock (start, end) {
+    return this.blobs.core.has(start, end)
+  }
+
+  async has (folder = '/') {
+    const entries = this.list(folder)
+    for await (const entry of entries) {
+      const b = entry.value.blob
+      const has = await this.hasBlock(b.blockOffset, b.blockOffset + b.blockLength)
+      if (!has) return false
+    }
+    return true
+  }
+
   // atm always recursive, but we should add some depth thing to it
   list (folder, opts = {}) {
     if (typeof folder === 'object') return this.list(undefined, folder)
--- a/index.js
+++ b/index.js
@@ -477,18 +477,28 @@ module.exports = class Hyperdrive extends ReadyResource {
     await Promise.allSettled(proms)
   }
 
-  hasBlock (start, end) {
+  _hasBlock (start, end) {
     return this.blobs.core.has(start, end)
   }
 
-  async has (folder = '/') {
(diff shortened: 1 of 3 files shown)


index.js
```diff
@@ -477,6 +477,20 @@ module.exports = class Hyperdrive extends ReadyResource {
     await Promise.allSettled(proms)
   }
 
+  hasBlock (start, end) {
```
mafintosh, line 480: dont make this public

index.js
```diff
@@ -477,6 +477,20 @@ module.exports = class Hyperdrive extends ReadyResource {
...
+    return this.blobs.core.has(start, end)
+  }
+
+  async has (folder = '/') {
```
mafintosh, line 484: you have to call blobs = await this.getBlobs(), otherwise its timing dependent

index.js
```diff
@@ -477,6 +477,20 @@ module.exports = class Hyperdrive extends ReadyResource {
...
+    const entries = this.list(folder)
+    for await (const entry of entries) {
+      const b = entry.value.blob
```
mafintosh, line 487: if (!b) continue

index.js
```diff
@@ -477,18 +477,28 @@ module.exports = class Hyperdrive extends ReadyResource {
...
-  async has (folder = '/') {
-    const entries = this.list(folder)
-    for await (const entry of entries) {
-      const b = entry.value.blob
-      const has = await this.hasBlock(b.blockOffset, b.blockOffset + b.blockLength)
-      if (!has) return false
+  async has (path = '/') {
+    if (!path) return false
+    await this.getBlobs()
+    const isFolder = path.endsWith('/')
```
mafintosh, line 487: this is too brittle, just do an .entry lookup to see if it has an entry, if no entry -> assume its a folder

index.js
```diff
@@ -477,18 +477,28 @@ module.exports = class Hyperdrive extends ReadyResource {
...
   }
 
-  async has (folder = '/') {
-    const entries = this.list(folder)
-    for await (const entry of entries) {
-      const b = entry.value.blob
-      const has = await this.hasBlock(b.blockOffset, b.blockOffset + b.blockLength)
-      if (!has) return false
+  async has (path = '/') {
+    if (!path) return false
```
mafintosh, line 485: remove this, up to the user

mafintosh:
12.1.0

PR #388, opened 2025-04-09
add function ignore

geordangesink:


index.js
```diff
@@ -500,8 +500,7 @@ module.exports = class Hyperdrive extends ReadyResource {
 
     folder = std(folder || '/', true)
 
-    const ignore = opts.ignore ? normalizeIgnore(opts.ignore) : null
-    const stream = opts && opts.recursive === false ? shallowReadStream(this.db, folder, false, ignore, opts) : this.entries(prefixRange(folder), { ...opts, ignore })
+    const stream = opts && opts.recursive === false ? shallowReadStream(this.db, folder, false, opts.ignore, opts) : this.entries(prefixRange(folder), { ...opts, ignore: opts.ignore })
```
mafintosh, line 503: const ignore = opts.ignore ? toIgnoreFunction(opts.ignore) : null

mafintosh: before this line and pass both places

index.js
```diff
@@ -758,12 +758,16 @@ async function getBlobsLength (db) {
...
-  return [].concat(ignore).map(e => unixPathResolve('/', e))
+function toIgnoreFunction (ignore) {
+  if (typeof ignore === 'function') return ignore
+
+  const all = [].concat(ignore).map(e => unixPathResolve('/', e))
+  return key => all.some(path => path === key || key.startsWith(path + '/'))
 }
 
 function isIgnored (key, ignore) {
-  return ignore.some(e => e === key || key.startsWith(e + '/'))
```
mafintosh, line 766: Ignore is always a function now so you can axe this whole method and just call ignore directly

PR #389, opened 2025-05-01
download can be done'ed or destroyed

rafapaezbas:


index.js
```diff
@@ -453,8 +448,8 @@ module.exports = class Hyperdrive extends ReadyResource {
       const b = entry.value.blob
       if (!b) return
       const blobs = await this.getBlobs()
-      await blobs.core.download({ start: b.blockOffset, length: b.blockLength }).downloaded()
-      return
+      const download = await blobs.core.download({ start: b.blockOffset, length: b.blockLength })
+      return new Download(download)
```
mafintosh, line 452: make it an array here so its always an array

lib/download.js
```diff
@@ -0,0 +1,13 @@
+module.exports = class Download {
+  constructor (coreDownloads = []) {
+    this._coreDownloads = Array.isArray(coreDownloads) ? coreDownloads : [coreDownloads]
+  }
+
+  destroy () {
+    this._coreDownloads.forEach((d) => d.destroy())
```
mafintosh, line 7: do a normal loop

PR #390, opened 2025-06-03
Fix race condition in tests using replication

HDegroote:
Relies on https://github.com/holepunchto/hypercore/pull/704

### DIFF
--- a/package.json
+++ b/package.json
@@ -24,7 +24,7 @@
   "dependencies": {
     "hyperbee": "^2.11.1",
     "hyperblobs": "^2.3.0",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -24,7 +24,7 @@
   "dependencies": {
     "hyperbee": "^2.11.1",
     "hyperblobs": "^2.3.0",
-    "hypercore": "^11.0.0",
+    "hypercore": "holepunchto/hypercore#update-min-length",
```
HDegroote, line 27: needs published version

PR #404, opened 2026-03-06
Prefetch db core before totalling length in `getBlobsLength()`

lejeunerenard:
Prefetching dramatically speeds up calculating the blobs length for large hyperdrives. It means more blocks are downloaded but the hyperbee blocks are relatively small and worth the tradeoff for speed.

The checkout needs to be `ready()`'ed because the length hasn't loaded yet by the time the prefetch is called.

### DIFF
--- a/index.js
+++ b/index.js
@@ -763,6 +763,10 @@ function generateContentManifest(m, key) {
 async function getBlobsLength(db) {
   let length = 0
 
+  // Prefetch db core
+  await db.ready() // Ensure length is loaded
+  await db.core.download({ start: 0, end: db.core.length }).done()
+
   for await (const { value } of db.createReadStream()) {
     const b = value && value.blob
     if (!b) continue
--- a/test.js
+++ b/test.js
@@ -11,6 +11,7 @@ const Hyperswarm = require('hyperswarm')
 const b4a = require('b4a')
 const getTmpDir = require('test-tmp')
 const unixPathResolve = require('unix-path-resolve')
+const safetyCatch = require('safety-catch')
 const Hyperdrive = require('./index.js')
 
 test('drive.core', async (t) => {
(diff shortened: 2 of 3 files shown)


lejeunerenard:
Added `debugging-stream` based test to make timing tests not flaky & the fix vs slow scenario more dramatically different in tests. We likely should consider splitting helpers into a separate file as there is a good chunk now.

test.js
```diff
@@ -1499,6 +1500,36 @@ test('getBlobsLength of empty drive', async (t) => {
   await corestore.close()
 })
 
+test('getBlobsLength large db - prefetch', async (t) => {
+  const store = new Corestore(await t.tmp())
+  t.teardown(() => store.close())
+  const a = new Hyperdrive(store.session())
+  t.teardown(() => a.close())
```
HDegroote: (detail) Unneeded (closed with the store)

index.js
```diff
@@ -763,6 +763,10 @@ function generateContentManifest(m, key) {
 async function getBlobsLength(db) {
   let length = 0
 
+  // Prefetch db core
+  await db.ready() // Ensure length is loaded
+  await db.core.download({ start: 0, end: db.core.length }).done()
```
HDegroote, line 768: Confirming: we make a trade-off here, correct? Trading a vastly increased speed of getting the blobs length for an increase in the amount of blocks we downloaded from the db: all vs a subset (since db.createReadStream() does not necessarily iterate over all blocks)

I think the trade-off makes sense, particularly if there's no other way to do a more targeted prefetch

HDegroote (approved):

index.js
lejeunerenard: Yep. That's the trade off we're making. To be more selective, we'd need to know without scanning the tree what blocks are required, which is exactly why this is slow (aka `createReadStream()` walks through the tree). Maybe this could be added to tree roots in the future, but doesn't exist currently. The b-tree is pretty small in general so while extra space isn't much. Also could be cleared later, though the next call for `getBlobsLength()` would download it again.

HDegroote: Agreed, thanks for the full explanation

test.js
lejeunerenard: Opted to close the drive instead of the store as the drive theoretically could have monitors as well as the two cores and isn't checked out. Not relevant for the test, but if only closing one and given tests are used as reference by others, then would pick the option that's slightly "better" in more cases.

HDegroote: Windows CI is failing, so I might have been wrong

HDegroote: Good point, so the correct pattern is to close all cores/drives opened from a corestore, as well as the store itself, correct?

lejeunerenard: I think it failed because the drive is using a session. I'm verifying via only closing the store now.

lejeunerenard: Looks like it was the sessions. So I'm thinking that reverting to closing both is the preferred pattern when the drive uses a session.

HDegroote: Sorry for the detour, we settled back on what you had originally. But at least I learned something from it

lejeunerenard: No worries. It was good to ask to keep tests simple & straightforward and I forgot that there was a corestore session otherwise the drive close alone would have been enough.

I've reverted the changes.

PR #406, opened 2026-03-18
Track whether `path` falling through `has()` had sub entries

lejeunerenard:
Previously it was assumed that if a `path` got to this point, the `path` was a directory (eg. `/dir`). But if there are no entries then it can't be asserted. In addition file paths (eg `/non-existent.txt`) would return true.

This adds a test for a non-existent `path` returning `false` and a fix by detecting if any entry was found before asserting `path` was a directory.

### DIFF
--- a/index.js
+++ b/index.js
@@ -475,13 +475,15 @@ module.exports = class Hyperdrive extends ReadyResource {
       if (!b) return false
       return await blobs.core.has(b.blockOffset, b.blockOffset + b.blockLength)
     }
+    let isDir = false
     for await (const entry of this.list(path)) {
+      isDir = true
       const b = entry.value.blob
       if (!b) continue
       const has = await blobs.core.has(b.blockOffset, b.blockOffset + b.blockLength)
       if (!has) return false
     }
-    return true
+    return isDir
   }
 
   // atm always recursive, but we should add some depth thing to it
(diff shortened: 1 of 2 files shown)


index.js
```diff
@@ -475,13 +475,15 @@ module.exports = class Hyperdrive extends ReadyResource {
       if (!b) return false
       return await blobs.core.has(b.blockOffset, b.blockOffset + b.blockLength)
     }
+    let isDir = false
     for await (const entry of this.list(path)) {
+      isDir = true
       const b = entry.value.blob
       if (!b) continue
```
HDegroote, line 482: Not a comment, just a question to double check I understand the logic: tThis triggers only when b is a directory, correct?

lejeunerenard: When `path` is a dir because it has entries under it when listing. The `this.list(path)` is to detect if the path is a directory since directories don't have explicit entries in the hyperbee.

test.js
```diff
@@ -838,6 +838,7 @@ test('drive.has(path)', async (t) => {
 
   t.absent(await mirror.drive.has('/parent/child/'))
   t.absent(await mirror.drive.has('/parent/child/grandchild2'))
+  t.absent(await mirror.drive.has('/non-existent.txt'), 'returns false for non-existent files')
```
HDegroote, line 841: Maybe add an additional assertion for non-existent dirs? Then all cases are tested

```
  t.absent(await mirror.drive.has('/non-existent/', 'returns false for non-existent directory'))

```

HDegroote (approved):
Nice find, fix looks good to me

PR #409, opened 2026-03-24
Fix flake by ensuring db length

nicolasfont:


lejeunerenard (approved):
Nice! This is a good temporary fix to ensure the db on the mirror is up to date. Since the test is verifying that the blob request can timeout correctly, ensuring the db length allows the assertion that the entry is available to pass.

The entire set of tests are based on faulty implicit timing assumptions so needs to be rewritten, but this fix should stop the tests from flaking in the mean time without creating a false positive.

PR #412, opened 2026-04-02
Rewrite Tests

danielchrist0pher:
Re-writing tests to fix inherent timing issues.

### DIFF
--- a/test.js
+++ b/test.js
@@ -412,15 +412,15 @@ test('watch(folder) basic', async function (t) {
   const watcher = drive.watch('/examples')
   await watcher.ready()
 
-  await drive.put('/b.txt', buf)
   const prevVersion = drive.version
+  await drive.put('/b.txt', buf)
   const next = watcher.next()
   await drive.put('/examples/b.txt', buf)
 
   const { value } = await next
   const [current, previous] = value
 
-  t.is(previous.version, prevVersion)
+  t.ok(previous.version !== prevVersion)
   t.is(await previous.get('/examples/b.txt'), null)
   t.alike(await current.get('/examples/b.txt'), buf)
 
@@ -840,6 +840,7 @@ test('drive.has(path)', async (t) => {
   t.absent(await mirror.drive.has('/non-existent/'), 'returns false for non-existent directory')
 
   await drive.put('/parent/sibling/grandchild1', nil)
+  await ensureDbLength(mirror.drive, drive.version)
 
   const downloadChild = mirror.drive.download('/parent/child/')
   await downloadChild.done()
@@ -955,9 +956,10 @@ test('drive.close() with openBlobsFromHeader waiting in the background', async (
   t.ok(drive.corestore.closed)
 })
 
-test.skip('drive.findingPeers()', async (t) => {
+test('drive.findingPeers()', async (t) => {
+  t.plan(2)
   const { drive, corestore, swarm, mirror } = await testenv(t)
-  await drive.put('/', b4a.from('/'))
+  await drive.put('/a', b4a.from('a'))
 
   swarm.on('connection', (conn) => corestore.replicate(conn))
   swarm.join(drive.discoveryKey, { server: true, client: false })
@@ -965,9 +967,17 @@ test.skip('drive.findingPeers()', async (t) => {
 
   mirror.swarm.on('connection', (conn) => mirror.corestore.replicate(conn))
   mirror.swarm.join(drive.discoveryKey, { server: false, client: true })
+
   const done = mirror.drive.findingPeers()
-  swarm.flush().then(done, done)
-  t.ok(await mirror.drive.get('/'))
+  const updating = mirror.drive.update({ wait: true })
+  try {
+    await Promise.all([waitForEvent(mirror.swarm, 'connection'), mirror.swarm.flush()])
+  } finally {
+    done()
+  }
+
+  t.ok(await updating)
+  t.alike(await mirror.drive.get('/a'), b4a.from('a'))
 })
 
 test('drive.mirror()', async (t) => {
@@ -1020,17 +1030,26 @@ test('drive.clear(path)', async (t) => {
   t.is(nowContent, null)
 })
 
-test.skip('drive.clear(path) with diff', async (t) => {
+test('drive.clear(path) with diff', async (t) => {
   const storage = await getTmpDir(t)
 
   const a = new Hyperdrive(new Corestore(storage))
+  t.teardown(() => a.close())
   await a.put('/file', b4a.alloc(4 * 1024))
+  const init = await a.get('/file', { wait: false })
+  t.ok(init)
+  const key = a.key
   await a.close()
 
-  const b = new Hyperdrive(new Corestore(storage))
+  const b = new Hyperdrive(new Corestore(storage), key)
+  t.teardown(() => b.close())
+  await b.getBlobs()
 
   const cleared = await b.clear('/file', { diff: true })
   t.ok(cleared.blocks > 0)
+
+  t.ok(await b.entry('/file'))
+  await t.exception(() => b.get('/file', { wait: false }), /BLOCK_NOT_AVAILABLE/)
 
   const cleared2 = await b.clear('/file', { diff: true })
   t.is(cleared2.blocks, 0)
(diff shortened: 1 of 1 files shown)


test.js
```diff
@@ -2057,6 +2051,14 @@ function replicateDebugStream(t, a, b, opts = {}) {
   return [s1, s2]
 }
 
-async function ensureDbLength(drive, length) {
-  await drive.checkout(length).db.core.get(length - 1)
+async function ensureDbLength(drive, length, timeout = 20000) {
+  await drive.checkout(length).db.core.get(length - 1, { timeout })
```
mafintosh:
```js
while (drive.db.core.length < length) {
  await new Promise(resolve => drive.db.core.once('append', resolve))
}
```

test.js
```diff
@@ -2057,6 +2051,14 @@ function replicateDebugStream(t, a, b, opts = {}) {
...
+}
+
+async function waitForAppendIfEmpty(core, timeout = 20000, message) {
+  if (core.length !== 0) return
+  await Promise.race([
+    once(core, 'append'),
+    new Promise((_, reject) => setTimeout(reject, timeout, new Error(message)))
```
mafintosh: you need to clear this timeout once the race finishes

test.js
lejeunerenard: Could also use `once()` aka:
```js
while (drive.db.core.length < length) await once(drive.db.core, 'append')
```

test.js
```diff
@@ -408,49 +410,41 @@ test('watch(folder) basic', async function (t) {
...
 
-  let next = watcher.next()
-  let onchange = null
-  next.then((data) => {
-    next = watcher.next()
-    onchange(data)
-  })
-
-  onchange = () => t.fail('should not trigger changes')
   await drive.put('/b.txt', buf)
```
lejeunerenard, line 416: This `.put()` is now "skipped". This part of the test was to verify that the `folder` passed to `.watch()` filters the files it watches. The new way the test is written does implicitly test this in that `.next()` is only called once, but IMHO it needs something more explicit. So maybe store the version before and showcase that the `previous` from the `watch()`'s `.next()` isn't equal, aka it skipped that update.

test.js
```diff
@@ -845,18 +839,20 @@ test('drive.has(path)', async (t) => {
 
   await drive.put('/parent/sibling/grandchild1', nil)
 
+  await ensureDbLength(mirror.drive, drive.version)
```
lejeunerenard, line 843: This new `await ensureDbLength()` is needed but the other two aren't from what I can tell. The first is needed to ensure that when the mirror is downloading it even knows all that it would need. The others aren't needed because the changes being tested are for the blobs which are unaffected by the db length since it doesnt change.

test.js
lejeunerenard: @9mzodiac your change to check the prevVersion before the `next()` call is slightly less implicit, but still leaves the question when reading the test, what the `await drive.put('/b.txt', buf)` is for. So I'd move `prevVersion` to above the `.put()` and check that it `t.not(previous.version, prevVersion)` to show that the `previous` doesn't include the `.put()`.

test.js
lejeunerenard: @9mzodiac you seem to have removed the one required `await ensureDbLength(mirror.drive, drive.version)` after `await drive.put('/parent/sibling/grandchild1', nil)`. The test flakes without it because of what I mentioned above.

danielchrist0pher: ahh good catch - I'll add that back

test.js
```diff
@@ -2057,6 +2117,45 @@ function replicateDebugStream(t, a, b, opts = {}) {
   return [s1, s2]
 }
 
-async function ensureDbLength(drive, length) {
-  await drive.checkout(length).db.core.get(length - 1)
+async function ensureDbLength(drive, length, timeout = 20000) {
```
lejeunerenard:
```suggestion
async function ensureDbLength(drive, length) {
```

test.js
```diff
@@ -1027,18 +1030,27 @@ test('drive.clear(path)', async (t) => {
   t.is(nowContent, null)
 })
 
-test.skip('drive.clear(path) with diff', async (t) => {
+test('drive.clear(path) with diff', async (t) => {
```
lejeunerenard: This test is flaking: https://github.com/holepunchto/hyperdrive/actions/runs/24104863312/job/70325422890?pr=412#step:3:459

This appears to be a bug in `hypercore` as the cleared blocks comes from there. So this test should be skipped for now.

I'm looking into fixing the underlying issue here: https://github.com/holepunchto/hypercore/pull/795

test.js
```diff
@@ -1070,20 +1082,30 @@ test('drive.clear(path) on a checkout', async (t) => {
   t.is(nowContent, null)
 })
 
-test.skip('drive.clearAll() with diff', async (t) => {
+test('drive.clearAll() with diff', async (t) => {
```
lejeunerenard: This test is flaking: https://github.com/holepunchto/hyperdrive/actions/runs/24104863312/job/70325422890?pr=412#step:3:485

This also appears to be a bug in `hypercore` as the cleared blocks comes from there. So this test should be skipped for now.

test.js
```diff
@@ -1520,19 +1542,19 @@ test('getBlobsLength large db - prefetch', async (t) => {
...
+  await b.checkout(targetVersion).db.core.get(targetVersion - 1, { timeout: 20000 })
 
   const bBlobsLength = await b.getBlobsLength()
   const end = Date.now()
 
   t.is(bBlobsLength, await a.getBlobsLength(), 'blob lengths match')
 
   t.comment('getBlobsLength() time in secs ' + (end - start) / 1000)
-  t.ok(end - start < 2_000, 'synced in a reasonable time')
 })
```
lejeunerenard, line 1557: Why was this test changed? Did it flake? Currently at ~4.6k runs locally without a flake.
The change is no longer testing the principle purpose of this test which was to verify that .getBlobsLength() was quick despite having a large number of files and a "slow" connection (10ms latency). Instead the rewrite only comments on the delta time and uses a larger timeout for getting the latest block.
In case this was the point of confusion, the append event is awaited only to ensure the peer  (b) has an updated length so when it queries the blobs length it checks the same version.

test.js
```diff
@@ -1772,46 +1795,87 @@ test.skip('upload/download can be monitored', async (t) => {
...
+
+  const getting = mirror.drive.get(file)
+
+  const sawUpload = (async () => {
+    if (!uploadDone()) await waitForEvent(uploadMonitor, 'update', uploadDone)
+  })()
+
+  const sawDownload = (async () => {
+    if (!downloadDone()) await waitForEvent(downloadMonitor, 'update', downloadDone)
+  })()
```
lejeunerenard: These are unnecessary, right? The `(down/up)loadDone()` predicate is testing your assertions later and if we get a file the `update` event should have already fired by the end. `upload` & `download` events from hypercore fire before the `.get()` completes.

lejeunerenard (changes requested):

test.js
lejeunerenard: Sorry I should clarify, I mean you don't need the `sawUpload` & `sawDownload` promises because you don't need `downloadDone` & `uploadDone` at all.
So you can remove these lines in addition to the lines this comment is for, etc:
https://github.com/holepunchto/hyperdrive/pull/412/changes/be9b83858c7eb50fb19f6cf35746059d074534c0#diff-13876b4beb64b9f156474dc78f9c923952a7ca210d4507b6b3135bbe244f8a60R1814-R1822

test.js
lejeunerenard: As discussed elsewhere, you were right to remove the timing delta assertion. Will post a better assertion using replicator stats as a suggestion.

test.js
```diff
@@ -1520,19 +1542,19 @@ test('getBlobsLength large db - prefetch', async (t) => {
...
-  replicateDebugStream(t, a, b, { latency: 10 })
-  await gotAppend
+  await b.checkout(targetVersion).db.core.get(targetVersion - 1, { timeout: 20000 })
 
   const bBlobsLength = await b.getBlobsLength()
   const end = Date.now()
 
   t.is(bBlobsLength, await a.getBlobsLength(), 'blob lengths match')
 
   t.comment('getBlobsLength() time in secs ' + (end - start) / 1000)
```
lejeunerenard:
```suggestion
  const num = 1_000
  for (let i = 0; i < num; i++) {
    await a.put('./file' + i, 'here')
  }

  const store2 = new Corestore(await t.tmp())
  t.teardown(() => store2.close())
  const b = new Hyperdrive(store2.session(), a.key)
  t.teardown(() => b.close())

  const gotAppend = once(b.core, 'append')
  replicateDebugStream(t, a, b, { latency: 10 })
  await gotAppend

  t.is(await b.getBlobsLength(), await a.getBlobsLength(), 'blob lengths match')

  t.comment('wireRequest sent', b.core.replicator.stats.wireRequest.tx)
  t.ok(b.core.replicator.stats.wireRequest.tx < 1.1 * num, 'synced within a reasonable amount of requests')
```

This is the replicator stat based way to assert the fix of prefetching the file structure for getting blob lengths.

lejeunerenard (approved):

PR #413, opened 2026-04-13
Refactor Tests

danielchrist0pher:
- Unit (test/unit.js) - no DHT / swarm
- Integration (test/integration.js) - replication without Hyperswarm.
- E2E (test/e2e.js) - uses Hyperswarm + hyperdht testnet


danielchrist0pher:
Got a base going - now going to refactor tests to use direct replication rather than hyperswarm.

danielchrist0pher:
Tests are refactored to use direct replication where possible. We do have one E2E test using hyperswarm - not sure this warrants its own dedicated "e2e" file. But nice to have, in case we want to grow the suite.

test/helpers.js
```diff
@@ -0,0 +1,237 @@
...
+
+async function streamToBuffer(stream) {
+  const chunks = []
+  for await (const chunk of stream) {
+    chunks.push(chunk)
+  }
+  return b4a.concat(chunks)
+}
+
+async function replicate(drive, swarm, mirror) {
```
lejeunerenard, line 122: This should be renamed something to do with swarming, maybe event just `swarm()`.

test/helpers.js
```diff
@@ -0,0 +1,237 @@
...
+  s1.pipe(s2).pipe(s1)
+
+  return [s1, s2]
+}
+
+async function ensureDbLength(drive, length) {
+  while (drive.db.core.length < length) await once(drive.db.core, 'append')
+}
+
+function pipeReplicate(drive, mirror) {
```
lejeunerenard, line 167: Rename to `replicate()` like other helpers eg https://github.com/holepunchto/hypercore/blob/6daacb9bcd1d6fb1ba203ade60db75a3d6027dae/test/helpers/index.js#L35

test/integration.js
lejeunerenard, line 1: All of these tests are unit tests.

lejeunerenard (changes requested):

PR #416, opened 2026-06-23
Fix download

rafapaezbas:
The Download class was broken for entries written with dedup: true. Dedup entries store blobs using a block map (an index of non-contiguous blocks) rather than a contiguous blockOffset/blockLength range. The previous code always downloaded using { start, length }, which is incorrect for dedup entries — it would request the wrong range from the blobs core, causing downloads to silently stall or fetch the wrong data.

The same bug affected has(), which also used the contiguous range form for all entries. Both are fixed via new _downloadEntry and _hasEntry helpers that branch on blob.blockMap, using resolved block indices for dedup entries and falling back to the range form otherwise. 

The same bug existed in downloadDiff. It manually constructed blob downloads using { start: b.blockOffset, length: b.blockLength }, always taking the contiguous range path and never handling dedup entries.

### DIFF
--- a/index.js
+++ b/index.js
@@ -484,16 +484,16 @@ module.exports = class Hyperdrive extends ReadyResource {
     const blobs = await this.getBlobs()
     const entry = !path || path.endsWith('/') ? null : await this.entry(path)
     if (entry) {
-      const b = entry.value.blob
-      if (!b) return false
-      return await blobs.core.has(b.blockOffset, b.blockOffset + b.blockLength)
+      const blob = entry.value.blob
+      if (!blob) return false
+      return await this._hasEntry(blobs, blob)
     }
     let isDir = false
     for await (const entry of this.list(path)) {
       isDir = true
-      const b = entry.value.blob
-      if (!b) continue
-      const has = await blobs.core.has(b.blockOffset, b.blockOffset + b.blockLength)
+      const blob = entry.value.blob
+      if (!blob) continue
+      const has = await this._hasEntry(blobs, blob)
       if (!has) return false
     }
     return isDir
@@ -656,6 +656,19 @@ module.exports = class Hyperdrive extends ReadyResource {
         ondrain = null
         cb(err)
       }
+    }
+  }
+
+  async _hasEntry(blobs, blob) {
+    if (blob.blockMap) {
+      const map = await blobs.getBlockMap(blob)
+      for (const block of map.blocks) {
+        const has = await blobs.core.has(block.index)
+        if (!has) return false
+      }
+      return true
+    } else {
+      return await blobs.core.has(blob.blockOffset, blob.blockOffset + blob.blockLength)
     }
   }
 
--- a/index.js
+++ b/index.js
@@ -441,10 +441,10 @@ module.exports = class Hyperdrive extends ReadyResource {
 
     for await (const entry of this.diff(length, folder, opts)) {
       if (!entry.left) continue
-      const b = entry.left.value.blob
-      if (!b) continue
+      const blob = entry.left.value.blob
+      if (!blob) continue
       const blobs = await this.getBlobs()
-      dls.push(blobs.core.download({ start: b.blockOffset, length: b.blockLength }))
+      dls.push(await Download.downloadEntry(blobs, blob))
     }
 
     return new Download(this, null, { downloads: dls })
--- a/index.js
+++ b/index.js
@@ -486,7 +486,7 @@ module.exports = class Hyperdrive extends ReadyResource {
     if (entry) {
       const blob = entry.value.blob
       if (!blob) return false
-      return await this._hasEntry(blobs, blob)
+      return this._hasEntry(blobs, blob)
     }
     let isDir = false
     for await (const entry of this.list(path)) {
@@ -668,7 +668,7 @@ module.exports = class Hyperdrive extends ReadyResource {
       }
       return true
     } else {
-      return await blobs.core.has(blob.blockOffset, blob.blockOffset + blob.blockLength)
+      return blobs.core.has(blob.blockOffset, blob.blockOffset + blob.blockLength)
     }
   }
 
--- a/index.js
+++ b/index.js
@@ -441,10 +441,10 @@ module.exports = class Hyperdrive extends ReadyResource {
 
     for await (const entry of this.diff(length, folder, opts)) {
       if (!entry.left) continue
-      const blob = entry.left.value.blob
-      if (!blob) continue
+      const b = entry.left.value.blob
+      if (!b) continue
       const blobs = await this.getBlobs()
-      dls.push(await Download.downloadEntry(blobs, blob))
+      dls.push(blobs.core.download({ start: b.blockOffset, length: b.blockLength }))
     }
 
     return new Download(this, null, { downloads: dls })
--- a/index.js
+++ b/index.js
@@ -661,6 +661,9 @@ module.exports = class Hyperdrive extends ReadyResource {
 
   async _hasEntry(blobs, blob) {
     if (blob.blockMap) {
+      const hasMap = await blobs.core.has(blob.blockOffset, blob.blockOffset + blob.blockLength)
+      if (!hasMap) return false
+
       const map = await blobs.getBlockMap(blob)
       for (const block of map.blocks) {
         const has = await blobs.core.has(block.index)
--- a/lib/download.js
+++ b/lib/download.js
@@ -22,9 +22,9 @@ module.exports = class Download extends ReadyResource {
         : await drive.entry(this.folder, this.options)
 
     if (entry) {
-      const b = entry.value.blob
-      if (!b) return
-      const download = await this._downloadEntry(blobs, b)
+      const blob = entry.value.blob
+      if (!blob) return
+      const download = await this._downloadEntry(blobs, blob)
       this.downloads.push(download)
       return
     }
--- a/lib/download.js
+++ b/lib/download.js
@@ -27,10 +27,12 @@ module.exports = class Download extends ReadyResource {
       const blob = entry.value.blob
       if (!blob) return
 
-      const mapDownload = await this._cancellableDownload(
-        this.constructor.downloadEntryMap(blobs, blob)
-      )
-      if (!mapDownload) return
+if (blob.blockMap) {
+  const mapDownload = await this._cancellableDownload(
+    this.constructor.downloadEntryMap(blobs, blob)
+  )
+  if (!mapDownload) return
+}
 
       const download = await this.constructor.downloadEntry(blobs, blob)
       this.downloads.push(download)
--- a/lib/download.js
+++ b/lib/download.js
@@ -27,12 +27,12 @@ module.exports = class Download extends ReadyResource {
       const blob = entry.value.blob
       if (!blob) return
 
-if (blob.blockMap) {
-  const mapDownload = await this._cancellableDownload(
-    this.constructor.downloadEntryMap(blobs, blob)
-  )
-  if (!mapDownload) return
-}
+      if (blob.blockMap) {
+        const mapDownload = await this._cancellableDownload(
+          this.constructor.downloadEntryMap(blobs, blob)
+        )
+        if (!mapDownload) return
+      }
 
       const download = await this.constructor.downloadEntry(blobs, blob)
       this.downloads.push(download)
--- a/lib/download.js
+++ b/lib/download.js
@@ -56,7 +56,7 @@ module.exports = class Download extends ReadyResource {
       }
 
       const results = await Promise.all(maps.map((m) => this._cancellableDownload(m)))
-      if (results.includes(false)) return // if true, it means the Download got cancelled
+      if (results.includes(false)) return
 
       for (const entry of entries) {
         const blob = entry.value.blob
@@ -72,7 +72,7 @@ module.exports = class Download extends ReadyResource {
 
   static async downloadEntry(blobs, blob) {
     if (blob.blockMap) {
-      const map = await blobs.getBlockMap(blob) // always called after downloadEntryMap, so its non-blocking
+      const map = await blobs.getBlockMap(blob)
       const blocks = []
       for (const block of map.blocks) blocks.push(block.index)
 
--- a/lib/download.js
+++ b/lib/download.js
@@ -8,7 +8,9 @@ module.exports = class Download extends ReadyResource {
     this.folder = folder
     this.options = options || {}
     this.downloads = this.options.downloads || []
-    this.destroyed = false
+    this._cancelled = new Promise((resolve) => {
+      this._cancel = resolve
+    })
     this.ready().catch(noop)
   }
 
@@ -24,6 +26,12 @@ module.exports = class Download extends ReadyResource {
     if (entry) {
       const blob = entry.value.blob
       if (!blob) return
+
+      const mapDownload = await this._cancellableDownload(
+        this.constructor.downloadEntryMap(blobs, blob)
+      )
+      if (!mapDownload) return
+
       const download = await this.constructor.downloadEntry(blobs, blob)
       this.downloads.push(download)
       return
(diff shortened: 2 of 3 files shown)


index.js
```diff
@@ -484,16 +484,16 @@ module.exports = class Hyperdrive extends ReadyResource {
     const blobs = await this.getBlobs()
     const entry = !path || path.endsWith('/') ? null : await this.entry(path)
     if (entry) {
-      const b = entry.value.blob
-      if (!b) return false
-      return await blobs.core.has(b.blockOffset, b.blockOffset + b.blockLength)
+      const blob = entry.value.blob
+      if (!blob) return false
+      return await this._hasEntry(blobs, blob)
```
lejeunerenard:
```suggestion
      return this._hasEntry(blobs, blob)
```
Awaiting adds another promise as it resolves the `_hasEntry` promise to only wrap that value in another promise.

index.js
```diff
@@ -659,6 +659,19 @@ module.exports = class Hyperdrive extends ReadyResource {
...
+  async _hasEntry(blobs, blob) {
+    if (blob.blockMap) {
+      const map = await blobs.getBlockMap(blob)
+      for (const block of map.blocks) {
+        const has = await blobs.core.has(block.index)
+        if (!has) return false
+      }
+      return true
+    } else {
+      return await blobs.core.has(blob.blockOffset, blob.blockOffset + blob.blockLength)
```
lejeunerenard:
```suggestion
      return blobs.core.has(blob.blockOffset, blob.blockOffset + blob.blockLength)
```
Same as above.

test.js
```diff
@@ -816,6 +816,70 @@ test('drive.downloadDiff(version, folder, [options])', async (t) => {
...
+  mirror.swarm.join(drive.discoveryKey, { server: false, client: true })
+  await mirror.swarm.flush()
+
+  const ws = await drive.createWriteStream('/entry', { dedup: true })
+  ws.write(Buffer.alloc(1024))
+  ws.end()
+
+  await ensureDbLength(mirror.drive, drive.version)
+
+  const download = await mirror.drive.download('/entry')
```
lejeunerenard:
```suggestion
  const download = mirror.drive.download('/entry')
```
`.download()` doesn't return a promise.

test.js
```diff
@@ -816,6 +816,70 @@ test('drive.downloadDiff(version, folder, [options])', async (t) => {
...
+    const ws = await drive.createWriteStream('/folder/entry', { dedup: true })
+    ws.write(Buffer.alloc(1024))
+    ws.end()
+  }
+
+  await drive.put('/folder/entry-b', Buffer.from('hello world'))
+
+  await ensureDbLength(mirror.drive, drive.version)
+
+  const download = await mirror.drive.download('/folder')
```
lejeunerenard:
```suggestion
  const download = mirror.drive.download('/folder')
```

test.js
```diff
@@ -855,6 +919,32 @@ test('drive.has(path)', async (t) => {
...
+  const { corestore, drive, swarm, mirror } = await testenv(t)
+  swarm.on('connection', (conn) => corestore.replicate(conn))
+  swarm.join(drive.discoveryKey, { server: true, client: false })
+  await swarm.flush()
+
+  mirror.swarm.on('connection', (conn) => mirror.corestore.replicate(conn))
+  mirror.swarm.join(drive.discoveryKey, { server: false, client: true })
+  await mirror.swarm.flush()
+
+  const ws = await drive.createWriteStream('/entry', { dedup: true })
```
lejeunerenard:
```suggestion
  const ws = drive.createWriteStream('/entry', { dedup: true })
```

test.js
```diff
@@ -816,6 +816,70 @@ test('drive.downloadDiff(version, folder, [options])', async (t) => {
...
+  swarm.on('connection', (conn) => corestore.replicate(conn))
+  swarm.join(drive.discoveryKey, { server: true, client: false })
+  await swarm.flush()
+
+  mirror.swarm.on('connection', (conn) => mirror.corestore.replicate(conn))
+  mirror.swarm.join(drive.discoveryKey, { server: false, client: true })
+  await mirror.swarm.flush()
+
+  {
+    const ws = await drive.createWriteStream('/folder/entry', { dedup: true })
```
lejeunerenard:
```suggestion
    const ws = drive.createWriteStream('/folder/entry', { dedup: true })
```

test.js
```diff
@@ -816,6 +816,70 @@ test('drive.downloadDiff(version, folder, [options])', async (t) => {
...
+  const { corestore, drive, swarm, mirror } = await testenv(t)
+  swarm.on('connection', (conn) => corestore.replicate(conn))
+  swarm.join(drive.discoveryKey, { server: true, client: false })
+  await swarm.flush()
+
+  mirror.swarm.on('connection', (conn) => mirror.corestore.replicate(conn))
+  mirror.swarm.join(drive.discoveryKey, { server: false, client: true })
+  await mirror.swarm.flush()
+
+  const ws = await drive.createWriteStream('/entry', { dedup: true })
```
lejeunerenard:
```suggestion
  const ws = drive.createWriteStream('/entry', { dedup: true })
```

test.js
```diff
@@ -816,6 +816,70 @@ test('drive.downloadDiff(version, folder, [options])', async (t) => {
...
+  await download.done()
+
+  const mirrorBlobs = await mirror.drive.getBlobs()
+  const driveBlobs = await drive.getBlobs()
+
+  const mirrorBlobsHash = await mirrorBlobs.core.treeHash()
+  const driveBlobsHash = await driveBlobs.core.treeHash()
+
+  t.is(mirrorBlobs.core.contiguousLength, driveBlobs.core.contiguousLength)
+  t.is(mirrorBlobsHash.toString('hex'), driveBlobsHash.toString('hex'))
```
lejeunerenard:
```suggestion
  t.alike(mirrorBlobsHash, driveBlobsHash, 'blob hashes match')
```

test.js
```diff
@@ -816,6 +816,70 @@ test('drive.downloadDiff(version, folder, [options])', async (t) => {
...
```
lejeunerenard:
```suggestion
  t.alike(mirrorBlobsHash, driveBlobsHash, 'blob hashes match')
```

test.js
```diff
@@ -855,6 +919,32 @@ test('drive.has(path)', async (t) => {
...
+  ws.write(Buffer.alloc(1024))
+  ws.end()
+
+  await ensureDbLength(mirror.drive, drive.version)
+  await mirror.drive.entry('/entry')
+
+  await mirror.drive.getBlobs()
+  await mirror.drive.blobs.core.get(1) // get map block
+
+  t.is(await mirror.drive.has('/entry'), false)
```
lejeunerenard:
```suggestion
  t.absent(await mirror.drive.has('/entry'), 'has() is false w/ map, but w/o blocks')
```

lejeunerenard (commented):
Great catch! Thanks for the PR!

Have some nits but should be good afterwards.

index.js
```diff
@@ -659,6 +659,19 @@ module.exports = class Hyperdrive extends ReadyResource {
     }
   }
 
+  async _hasEntry(blobs, blob) {
+    if (blob.blockMap) {
+      const map = await blobs.getBlockMap(blob)
```
marcus-pousette-hp, line 667: Docs says "has(...)" is a local check. https://github.com/holepunchto/hyperdrive#await-drivehaspath 

but here we seem to wait/download the blockmap from peers. I think this is non ideal. 

We could add a early return like 

```ts
const hasMap = await blobs.core.has(blob.blockOffset, blob.blockOffset + blob.blockLength)
  if (!hasMap) return false
```

so that the test below passes


```ts
test('drive.has dedup entry does not download block map', async (t) => {
  t.plan(3)

  const { corestore, drive, mirror } = await testenv(t)

  const s1 = corestore.replicate(true)
  const s2 = mirror.corestore.replicate(false)
  s1.pipe(s2).pipe(s1)

  t.teardown(() => {
    s1.destroy()
    s2.destroy()
  })

  const ws = drive.createWriteStream('/entry', { dedup: true })
  const done = new Promise((resolve, reject) => {
    ws.once('error', reject)
    ws.once('finish', resolve)
  })

  ws.write(Buffer.alloc(1024))
  ws.write(Buffer.alloc(1024))
  ws.end()
  await done

  await ensureDbLength(mirror.drive, drive.version)

  const entry = await mirror.drive.entry('/entry')
  const blob = entry.value.blob
  const blobs = await mirror.drive.getBlobs()

  t.absent(
    await blobs.core.has(blob.blockOffset, blob.blockOffset + blob.blockLength),
    'sanity: block map is not local before has()'
  )

  let downloads = 0
  blobs.core.on('download', ondownload)
  t.teardown(() => blobs.core.off('download', ondownload))

  t.absent(await mirror.drive.has('/entry'), 'entry data is not local')
  t.is(downloads, 0, 'has() should not download while checking local state')

  function ondownload() {
    downloads++
  }
})
```

index.js
marcus-pousette-hp, line 447: Seems like this path also is suffering from the same problem.  Can this also be fixed part of this PR?


```ts
test('drive.downloadDiff downloads dedup entry blocks', async (t) => {
  t.plan(1)

  const { corestore, drive, mirror } = await testenv(t)

  const s1 = corestore.replicate(true)
  const s2 = mirror.corestore.replicate(false)
  s1.pipe(s2).pipe(s1)

  t.teardown(() => {
    s1.destroy()
    s2.destroy()
  })

  const version = drive.version
  const expected = Buffer.alloc(1024, 7)

  const ws = drive.createWriteStream('/entry', { dedup: true })
  const done = new Promise((resolve, reject) => {
    ws.once('error', reject)
    ws.once('finish', resolve)
  })

  ws.write(expected)
  ws.end()
  await done

  await ensureDbLength(mirror.drive, drive.version)

  const download = await mirror.drive.downloadDiff(version, '/')
  await download.done()

  t.alike(await mirror.drive.get('/entry', { wait: false }), expected)
})
```

marcus-pousette-hp:
See my point above. Review was co-performed with AI. Other than these problems I did not find anything else meaningful to adjust

index.js
rafapaezbas: added

marcus-pousette-hp:
Downloads does not seems to be cancellable anymore if DB entry is available but the block-map block download().destroy() can hang, and downloadDiff() may not return a Download handle at all.

Can you try the test below?

Disclaimer co-written with AI

```ts
test('dedup download can be destroyed while block map is unavailable', async (t) => {
  t.plan(1)

  const { corestore, drive, mirror } = await testenv(t)

  const s1 = corestore.replicate(true)
  const s2 = mirror.corestore.replicate(false)
  s1.pipe(s2).pipe(s1)

  const ws = drive.createWriteStream('/entry', { dedup: true })
  ws.write(Buffer.alloc(1024))
  ws.write(Buffer.alloc(1024))
  ws.end()

  await new Promise((resolve, reject) => {
    ws.once('error', reject)
    ws.once('finish', resolve)
  })

  await ensureDbLength(mirror.drive, drive.version)
  await mirror.drive.entry('/entry')

  s1.destroy()
  s2.destroy()

  const download = mirror.drive.download('/entry')
  download.destroy()

  await download.close()
  t.pass('download closed')
})
```

downloadDiff would regress on a similiar test.

marcus-pousette-hp:
So download seems to be fixed now! But downloadDiff seems to still suffer with the cancellability. 

```ts

test('dedup downloadDiff returns while block map is unavailable', async (t) => {
  t.plan(1)

  const { corestore, drive, mirror } = await testenv(t)

  const s1 = corestore.replicate(true)
  const s2 = mirror.corestore.replicate(false)
  s1.pipe(s2).pipe(s1)

  const version = drive.version

  const ws = drive.createWriteStream('/entry', { dedup: true })
  ws.write(Buffer.alloc(1024))
  ws.write(Buffer.alloc(1024))
  ws.end()

  await new Promise((resolve, reject) => {
    ws.once('error', reject)
    ws.once('finish', resolve)
  })

  await ensureDbLength(mirror.drive, drive.version)
  await mirror.drive.entry('/entry')

  s1.destroy()
  s2.destroy()

  const download = await mirror.drive.downloadDiff(version, '/')
  download.destroy()

  t.pass('downloadDiff returned a cancellable handle')
})

```

in practice I think 

await downloadDiff()
  OK to wait for: db diff scan, creating download ranges
  Not OK to wait forever for: dedup map/data blob availability

await download.done()
  OK to wait for: map blocks and actual data blocks

Disclaimer test cowritten with AI

index.js
```diff
@@ -441,10 +441,10 @@ module.exports = class Hyperdrive extends ReadyResource {
 
     for await (const entry of this.diff(length, folder, opts)) {
       if (!entry.left) continue
-      const b = entry.left.value.blob
-      if (!b) continue
+      const blob = entry.left.value.blob
+      if (!blob) continue
       const blobs = await this.getBlobs()
-      dls.push(blobs.core.download({ start: b.blockOffset, length: b.blockLength }))
+      dls.push(await Download.downloadEntry(blobs, blob))
```
marcus-pousette-hp: Move this into the stacked PR

test.js
```diff
@@ -816,6 +816,159 @@ test('drive.downloadDiff(version, folder, [options])', async (t) => {
...
+  s2.destroy()
+
+  const download = mirror.drive.download('/folder')
+  download.destroy()
+  await download.close()
+
+  t.pass('folder download closed cleanly')
+})
+
+test('drive.downloadDiff dedup entry', async (t) => {
```
marcus-pousette-hp: Move this test too into the same stacked PR

lib/download.js
```diff
@@ -8,25 +8,31 @@ module.exports = class Download extends ReadyResource {
...
-        start: b.blockOffset,
-        length: b.blockLength
-      })
+      const blob = entry.value.blob
+      if (!blob) return
+
+      const mapDownload = await this._cancellableDownload(
+        this.constructor.downloadEntryMap(blobs, blob)
+      )
+      if (!mapDownload) return
```
marcus-pousette-hp:
```suggestion
if (blob.blockMap) {
  const mapDownload = await this._cancellableDownload(
    this.constructor.downloadEntryMap(blobs, blob)
  )
  if (!mapDownload) return
}
```

should this predownload be guarded with `if (blob.blockMap)`? For normal blobs this waits for the file data during `_open()`, changing the old behavior

marcus-pousette-hp (dismissed):

test.js
```diff
@@ -855,6 +975,80 @@ test('drive.has(path)', async (t) => {
...
+  ws.write(Buffer.alloc(1024))
+  ws.write(Buffer.alloc(1024))
+  ws.write(Buffer.alloc(1024))
+  ws.end()
+
+  await ensureDbLength(mirror.drive, drive.version)
+  await mirror.drive.entry('/entry')
+
+  await mirror.drive.getBlobs()
+  await mirror.drive.blobs.core.get(1) // get map block
```
lejeunerenard:
```suggestion
  const entry = await mirror.drive.entry('/entry')

  await mirror.drive.getBlobs()
  await mirror.drive.blobs.core.get(entry.value.blob.blockOffset) // get map block
```
Using the entry's blob metadata makes it clear that the block index will be correct and will adjust for any changes to the test. A hard coded `1` requires trusting the author's original intent is preserved and the test didn't change in a way that `1` leads to a false positive.

lejeunerenard (commented):
Single nit but otherwise looks good assuming that `downloadDiff` will be fixed in #418

test.js
```diff
@@ -816,6 +816,126 @@ test('drive.downloadDiff(version, folder, [options])', async (t) => {
...
+  swarm.join(drive.discoveryKey, { server: true, client: false })
+  await swarm.flush()
+
+  mirror.swarm.on('connection', (conn) => mirror.corestore.replicate(conn))
+  mirror.swarm.join(drive.discoveryKey, { server: false, client: true })
+  await mirror.swarm.flush()
+
+  const ws = drive.createWriteStream('/entry', { dedup: true })
+  ws.write(Buffer.alloc(1024))
+  ws.end()
```
lejeunerenard, line 888: Need to await `finish` here too.

rafapaezbas: added

lejeunerenard (approved):

PR #420, opened 2026-08-19
Download diff

rafapaezbas:


lib/download.js
```diff
@@ -8,16 +8,40 @@ module.exports = class Download extends ReadyResource {
...
+
+  _destroyAll() {
+    if (this._stream) this._stream.destroy()
+    for (const m of this._maps) m.destroy()
+    for (const d of this.downloads) d.destroy()
+  }
+
+  async _download() {
     const drive = this.drive
     const blobs = await drive.getBlobs()
```
marcus-pousette-hp: Cancelability should be fixed. See this test that I think should pass

```ts 
test('downloadDiff destroy does not wait for blobs', async (t) => {
  const { drive, mirror } = await testenv(t)

  // Do not replicate, so the mirror cannot open its blob store.
  const download = await mirror.drive.downloadDiff(drive.version, '/')
  const closed = once(download, 'close')

  download.destroy()

  let timer = null
  const didClose = await Promise.race([
    closed.then(() => true),
    new Promise((resolve) => {
      timer = setTimeout(() => resolve(false), 250)
    })
  ])

  clearTimeout(timer)
  t.ok(didClose, 'download closes without waiting for blobs')

  // Unblock the broken implementation so the test exits cleanly.
  await mirror.drive.close()
  await closed
})
```

index.js
```diff
@@ -437,17 +437,7 @@ module.exports = class Hyperdrive extends ReadyResource {
...
-    for await (const entry of this.diff(length, folder, opts)) {
-      if (!entry.left) continue
-      const b = entry.left.value.blob
-      if (!b) continue
-      const blobs = await this.getBlobs()
-      dls.push(blobs.core.download({ start: b.blockOffset, length: b.blockLength }))
-    }
-
-    return new Download(this, null, { downloads: dls })
+    return new Download(this, folder, { ...opts, diff: length })
```
marcus-pousette-hp, line 444: For this change the API will behave differently 

```ts
diff(version, folder, opts) // works
diff(version, opts)         // works

downloadDiff(version, folder, opts) // works
downloadDiff(version, opts)         // broken
```

should we fix downloadDiff so it behaves the same as before the PR? (from the usage, or is the legacy behaivour you wish to prune?)

lib/download.js
```diff
@@ -68,6 +94,35 @@ module.exports = class Download extends ReadyResource {
...
+      }
+
+      blobsDiff.push(blob)
+    }
+
+    const results = await Promise.all(this._maps.map((m) => this._cancellableDownload(m)))
+    if (results.includes(false)) return
+
+    for (const blob of blobsDiff) {
+      const download = await this.constructor.downloadEntry(blobs, blob)
```
marcus-pousette-hp, line 123: We should check blob.blockLength and return early if 0 here. Else the test below will fail (which I think should pass) (please add this regression). 

I am not sure yet what the ideal solution is here, but the test below showcases the problem. Also ordinary donwload might also suffer and one solution could maybe fix both

```ts 
test('downloadDiff handles an empty dedup entry', async (t) => {
  const { drive } = await testenv(t)
  const version = drive.version

  const ws = drive.createWriteStream('/empty', { dedup: true })
  ws.end()
  await once(ws, 'finish')

  const download = await drive.downloadDiff(version, '/')

  await t.execution(
    download.done(),
    'an empty dedup entry requires no blocks'
  )
})
```

marcus-pousette-hp (changes requested):
Thanks for the PR! I made one review pass now. For the API changes I don't have sufficient information yet to say what is best but I guess preserving behavior is the way to go. 

Co-authored with AI

index.js
rafapaezbas: fixed + test

lib/download.js
rafapaezbas: fixed + added test. Malformed blobs still would need some fixes in `has` and `get` methods but imho it should be a different PR

lib/download.js
rafapaezbas: added test (required Promise.race for getBlobs when starting the download)

lib/download.js
```diff
@@ -8,15 +8,41 @@ module.exports = class Download extends ReadyResource {
...
+    if (this._stream) this._stream.destroy()
+    for (const m of this._maps) m.destroy()
+    for (const d of this.downloads) d.destroy()
+  }
+
+  async _download() {
     const drive = this.drive
-    const blobs = await drive.getBlobs()
+
+    const blobs = await Promise.race([drive.getBlobs(), this._cancelled])
```
marcus-pousette-hp, line 42: destroy() is fixed by this wait still bypasses the diff options. See the test below that I think should pass 

```ts
test('downloadDiff timeout does not wait for blobs', async (t) => {
  const { drive, mirror } = await testenv(t)

  await drive.put('/file', Buffer.from('data'))

  // Deliberately do not replicate.
  const download = await mirror.drive.downloadDiff(drive.version, '/', {
    timeout: 50
  })

  let timer = null
  const error = await Promise.race([
    download.done().then(
      () => null,
      (error) => error
    ),
    new Promise((resolve) => {
      timer = setTimeout(() => resolve(null), 250)
    })
  ])

  clearTimeout(timer)
  t.is(error && error.code, 'REQUEST_TIMEOUT')

  // Let the currently broken implementation exit cleanly.
  download.destroy()
  await download.close()
})
```

marcus-pousette-hp (changes requested):
Thanks for the update, one cancellation issue seems to be left

lib/download.js
rafapaezbas: timeout option is not documented for downloadDiff so I am not entirely sure how supported it is in main since it isn't documented in hyperbee createDiffStream either. 


Some options here:

1. We can add a timeout to the promise race.
2. Add a timeout in getBlobs method.

Anyway I don't think this is a regression of the PR (notice that getBlobs never took a timeout argument) so imho it could be fixed in a different PR.

marcus-pousette-hp: Good point. I looked through again and initially also thought about reordering so getBlobs would be called later, but that would only shield from of the failure mode a little bit. I reconsider here and I think we should do a follow up PR instead later to add timeout, wait: false, and signal correctly/consistently (in the blob opening) and not pretend that a half working solution is good enough for these options.

marcus-pousette-hp (dismissed):

lib/download.js
```diff
@@ -68,6 +96,35 @@ module.exports = class Download extends ReadyResource {
...
+
+    const blobsDiff = []
+
+    for await (const entry of this._stream) {
+      if (!entry.left) continue
+
+      const blob = entry.left.value.blob
+      if (!blob || !blob.blockLength) continue
+
+      if (blob.blockMap && blob.blockLength) {
```
lejeunerenard, line 112:
```suggestion
      if (blob.blockMap) {
```

The `blockLength` check is redundant with the check above.

lib/download.js
```diff
@@ -46,18 +73,19 @@ module.exports = class Download extends ReadyResource {
         // ignore
       }
 
-      const maps = []
       const entries = []
       for await (const entry of drive.list(this.folder, this.options)) {
         const blob = entry.value.blob
-        if (!blob) continue
+        if (!blob || !blob.blockLength) continue
         if (blob.blockMap && blob.blockLength) {
```
lejeunerenard, line 80:
```suggestion
        if (blob.blockMap) {
```

`blockLength` check is redundant.

lib/download.js
```diff
@@ -46,18 +73,19 @@ module.exports = class Download extends ReadyResource {
...
-          maps.push(blobs.core.download({ start: blob.blockOffset, length: blob.blockLength }))
+          this._maps.push(
+            blobs.core.download({ start: blob.blockOffset, length: blob.blockLength })
```
lejeunerenard, line 82:
```suggestion
            this.constructor.downloadEntryMap(blobs, blob)
```

Not exactly relevant to the PR, but since it's touched, makes sense to change.

lib/download.js
```diff
@@ -68,6 +96,35 @@ module.exports = class Download extends ReadyResource {
...
+
+      blobsDiff.push(blob)
+    }
+
+    const results = await Promise.all(this._maps.map((m) => this._cancellableDownload(m)))
+    if (results.includes(false)) return
+
+    for (const blob of blobsDiff) {
+      const download = await this.constructor.downloadEntry(blobs, blob)
+      this.downloads.push(download)
```
lejeunerenard, line 124: Downloads aren't cancellable as well correct? If the download is cancelled mid processing this loop, it would continue downloading and new download requests wouldn't be tracked etc.

This would also apply to `_download()`.

rafapaezbas: when a download is cancelled, hyperdrive iterates the downloads collection calling destroy on every instance https://github.com/holepunchto/hypercore/blob/main/lib/download.js#L46-L48

lejeunerenard: Yes but my point is that you can get into a situation where destroy isn't called on a hypercore download if the hyperdrive download destroy is called mid processing this loop. Because this loop doesn't check for whether its destroyed or cancelled it can continue adding downloads.

Here's a test for `downloadDiff()`:
```js
test(
  'downloadDiff destroy mid the per-blob download loop stops it from queuing more downloads',
  async (t) => {
    const drive = new Hyperdrive(new Corestore(await t.tmp()))
    t.teardown(() => drive.close())
    await drive.ready()

    const mirror = new Hyperdrive(new Corestore(await t.tmp()), drive.key)
    t.teardown(() => mirror.close())

    const num = 50

    let finishes = []
    for (let i = 0; i < num; i++) {
      const name = '/file' + i
      const ws = drive.createWriteStream(name, { dedup: true })
      ws.write(b4a.alloc(1024, name))
      ws.write(b4a.alloc(1024, name))
      finishes.push(once(ws, 'finish'))
      ws.end()
    }
    await Promise.all(finishes)

    const version = drive.version

    t.comment('replicating')
    const streams = replicateDirect(t, drive, mirror, { teardown: false })

    // Ensure one download is done and will be faster than network, needs to be first
    await mirror.download('/file0').done()
    for (const s of streams) s.destroy()

    await new Promise(setImmediate) // Not 100% sure its needed

    t.comment('create difference')
    finishes = []
    for (let i = 0; i < num; i++) {
      const name = '/file' + i
      const ws = drive.createWriteStream(name, { dedup: true })
      ws.write(b4a.alloc(1024, 'new-' + name))
      ws.write(b4a.alloc(1024, 'new-' + name))
      finishes.push(once(ws, 'finish'))
      ws.end()
    }
    await Promise.all(finishes)

    t.comment('re-replicate w/ diff')
    replicateDirect(t, drive, mirror)
    await ensureDbLength(mirror, drive.version)

    const diffDownload = await mirror.downloadDiff(version, '/')
    const blobs = await mirror.getBlobs()

    const haveADownload = new Promise((resolve) => {
      blobs.core.on('download', () => {
        // Ensure downloading block map results in a download entry
        if (diffDownload.downloads.length) resolve()
      })
    })
    await haveADownload

    t.not(diffDownload.downloads.length, num, 'Doesnt have all downloads yet')
    t.ok(
      diffDownload.downloads.some((d) => d.request.context !== null),
      'some requests are in progress'
    )

    const closed = once(diffDownload, 'close')
    diffDownload.destroy()

    // Allow map request still in flight to end
    await Promise.all(diffDownload._maps.map((d) => d.done()))
    const postDestroyDownloadNumber = diffDownload.downloads.length

    await closed

    t.absent(
      diffDownload.downloads.some((d) => d.request.context !== null),
      'destroy() stops all existing requests'
    )

    t.is(diffDownload.downloads.length, postDestroyDownloadNumber)

    t.absent(
      diffDownload.downloads.some((d) => d.request.context !== null),
      'all downloads still stopped'
    )
  }
)
```

The fix would be to add `if (this._isCancelled()) return download.destroy()` after each `const download = await this.constructor.downloadEntry(blobs, blob)` entry.

rafapaezbas: added

lejeunerenard (approved):

